PR #20, opened 2022-11-29
Update version of IJ and CfD

rock3r:
Since the CfD plugin is out of date, but we need the newer version (and to support newer IJ versions), this smol PR bumps the numbers where needed and replaces the usage of the CfD plugin with a direct Maven dependency*.

We also bump the Java version to 17, matching the newer IJ versions' JBR.

--

* As pointed out by @igordmn, this is an acceptable workaround until a more permanent solution is found to support CfD in IJP. The only downside is that the entire runtime will be bundled with each plugin, but that's not an issue for now.

### DIFF
--- a/themes/intellij/idea/build.gradle.kts
+++ b/themes/intellij/idea/build.gradle.kts
@@ -10,7 +10,7 @@ kotlin {
     target {
         compilations.all {
             kotlinOptions {
-                jvmTarget = "11"
+                jvmTarget = "17"
                 freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn", "-Xopt-in=kotlinx.coroutines.ExperimentalCoroutinesApi")
             }
         }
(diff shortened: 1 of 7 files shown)


themes/intellij/idea/build.gradle.kts
```diff
@@ -38,7 +38,7 @@ repositories {
 }
 
 dependencies {
-    compileOnly(compose.desktop.currentOs) {
+    implementation(compose.desktop.currentOs) {
```
igordmn, line 41: When we start publishing Jewel as a Maven artifact, this need to be replaced by this:
```
    implementation(compose.desktop.linux_arm64) { ... }
    implementation(compose.desktop.linux_x64) { ... }
    implementation(compose.desktop.windows_x64) { ... }
    implementation(compose.desktop.macos_arm64) { ... }
    implementation(compose.desktop.macos_x64) { ... }
```

igordmn (approved):

themes/intellij/idea/build.gradle.kts
rock3r: Sure thing. Please let me know when that's the case so I can update here!

rock3r: Will we be losing the nice notation? Can't we keep it and just change the underlying implementation? Cause I don't expect I'll want to manually declare all the platforms...

igordmn: > Will we be losing the nice notation

For that we will probably need something like `compose.desktop.all`, I will add it next time I change this piece of code.

lamba92 (approved):

PR #21, opened 2022-12-05
Reorganise project structure

rock3r:
This cleans up the project structure so that @devkanro can start bringing in the code from https://github.com/ButterCam/compose-jetbrains-theme

There is still some issues with running the standalone sample for whatever reason:

```
Circular dependency between the following tasks:
:samples:standalone:classes
\--- :samples:standalone:compileJava
     +--- :samples:standalone:compileKotlin
     |    \--- :samples:standalone:jar
     |         +--- :samples:standalone:classes (*)
     |         +--- :samples:standalone:compileKotlin (*)
     |         \--- :samples:standalone:inspectClassesForKotlinIC
     |              +--- :samples:standalone:classes (*)
     |              \--- :samples:standalone:compileKotlin (*)
     \--- :samples:standalone:jar (*)
```

I don't know what's going on there, so any help would be appreciated @lamba92.

The IDE sample does work, at least.

The new structure looks like this:

image

Where the `new-ui` theme is empty and just a placeholder for @devkanro's impl.


devkanro:
The reason why the standalone project cannot be run is that when gradle resolves dependencies, if the names of the two sub-projects are the same, unexpected problems will occur.

Please avoid two subprojects with the same folder name.

Reference: 
- gradle/gradle#16986

rock3r:
Got it, thanks @devkanro

rock3r:
All fixed, and I have also added a new wonderful standalone sample app :D

image

PR #26, opened 2022-12-14
Convention plugins & add ktlint

### LINKED ISSUE #22: Fix up repositories setup in buildscripts
For whatever reason, the `repositories` block we set in `settings.gradle.kts` seems not to be applied to all projects.

This should be fixed, and all the various `repositories` block anywhere else removed.

### LINKED ISSUE #23: Create configuration plugin
We should get rid of the `allprojects` block in the root `build.gradle.kts` and replace it with a plugin so that we can simplify and improve our common module setup, while at the same time improving configuration cacheability.

### LINKED ISSUE #24: Determine if we still need the Compose compiler version override
In the root buildscript, we have:

```kotlin
configurations.all {
    resolutionStrategy.dependencySubstitution {
        substitute(module("org.jetbrains.compose.compiler:compiler")).apply {
            using(module("androidx.compose.compiler:compiler:1.2.1-dev-k1.7.10-27cf0868d10"))
        }
    }
}
```

Is this still required? Shouldn't be!

rock3r:
Rewrites the Gradle setup to use convention plugins instead of repetition. Also adds ktlint to the project.

Fixes #22
Fixes #23
Fixes #24


rock3r:
Looks like sarif parsing is having issues :/

lamba92 (approved):

PR #27, opened 2022-12-21
New UI refactoring

devkanro:
See Issue #28 for more information


rock3r:
@fscarponi FYI — you may want to check this out while you work on the uber refactoring to prep for the bridge

devkanro:
I will be closed this because we adopt new pattern to impl control style

PR #30, opened 2022-12-22
Avoid putting compose content in condition branches

devkanro:
When switch the `isDark` argument, the `content` will be full recomposed, because content in different branch is different compose scope, remembers also will be re-evaluated.


rock3r (approved):
Ace!

PR #33, opened 2022-12-28
Fix button focus style

devkanro:
- Make pressed same as normal state style
- Draw halo border as outer style
- Fix double focusable

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/jewel/components/Button.kt
+++ b/core/src/main/kotlin/org/jetbrains/jewel/components/Button.kt
@@ -3,8 +3,8 @@
 package org.jetbrains.jewel.components
 
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.focusable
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/jewel/components/Button.kt
```diff
@@ -158,14 +148,9 @@ fun Button(
...
             )
-            .focusable(
-                enabled = enabled && focusable,
-                interactionSource = interactionSource
-            )
-            .then(pointerModifier)
             .then(shapeModifier)
             .then(haloModifier)
-            .clip(appearance.shape),
+        /*.clip(appearance.shape)*/,
```
rock3r, line 153: If we don't need this, can we remove it?

devkanro: Removed

rock3r (approved):

Issue #28, opened 2022-12-21
Refactoring New UI implementation based on core components


devkanro:
## Look and feel
- [x] **Palettes** Basic colors have been added, it will be changed in future
- [x] **Painters** SVGs have been redesigned for Checkbox and Radio button, it does not copy from IJ because there are many temp icons in 22.3
- [ ] **Metrics** Just copy from Darcula theme, need check and adjust
- [x] **Typography** Change the default font to Inter

## Core component support
- [ ] **Focus style support** 
  - #34 

## Consistency check
- [x] **Check box** 
- [x] **Radio button**
  - #31
  - #32
- [x] **Buttons** Need focus style support
  - #33
- [ ] **Slider** Current implemented by swing component
- [x] **TextField** Missing round border style, focus style, heading icon and trailer icon
  - #43
  - #27
- [x] **ProgressBar** Missing implementation
  - #27
- [ ] **SegmentedButtonColors** Missing implementation
- [x] **TextArea** Missing implementation
  - #39
- [x] **Link and DropdownLink** Missing implementation
- [ ] **ToolTip** Missing implementation
- [x] **DropdownMenu** Missing implementation
- [ ] **ContextMenu** Missing implementation
- [ ] **ComboBox** Missing implementation
- [ ] **And more...**

## Consistency Issue
- #35 

## Project struct
- [ ] **Remove old impl**
- [ ] **Grallery app**

RivanParmar:
@devkanro Are contributions allowed for this issue? If yes, then I would like to work on it.

devkanro:
@RivanParmar  Of course, I'm now blocked by some other works, feel free to submit PRs, but don't forget to check the [manual](https://github.com/JetBrains/jewel/blob/main/CONTRIBUTING.md)

RivanParmar:
@devkanro Are there any guidelines on how to structure the code? Since `Checkbox`, `RadioButton` and others are structured differently than each other.

rock3r:
The guiding principle is to try and do as close to Material as possible, because that API has been iterated and battle tested a lot, and also folks are familiar with it.

And yes, this means we'll have to go back and tweak existing code in this repo...

RivanParmar:
@rock3r What is the difference between a TextField and a TextArea? If it's same except for the number of lines, then would it be okay to implement TextField and TextArea in the same file with a common base?

rock3r:
[...]

PR #39, opened 2023-04-09
Add TextArea

RivanParmar:
Related issue: #28 

The `TextArea` has been implemented similar to the `TextField` as they both share many things in common. The only differences being that a `TextArea` supports multi-line input and does not show a border around itself when it gets focus.


rock3r:
@fscarponi assigning to you since you'll be doing work in this area

fscarponi:
From a quick check, it seems okay. I will proceed with the merge to avoid the incoming refactoring from falling too far behind the branch

fscarponi (approved):
From a quick check, it seems okay. I will proceed with the merge to avoid the incoming refactoring from falling too far behind the branch.

PR #40, opened 2023-04-10
Update project link in sample

pan3793:


devkanro:
Thanks for your contribution, the current implementation of the new UI is directly copied from the ButterCam/compose-jetbrains-them project.

We're working on cleaning up and refactoring this code to blend in with jewel's original code.

In the meantime, we recommend that you do not use this project for production.

PR #42, opened 2023-04-26
Selectable Lazy Column, Tree

fscarponi:
**OBSOLETE: SCROLL TO NEXT COMMENT**

New components (`SelectableLazyColumn`, `Tree`) can be found in the `foundation` package.

Int UI Demo

Those elements have a default implementation of `KeyBindingScopedActions`, and support focus and selection.

Standalone Int UI sample `org/jetbrains/jewel/samples/standalone/expui/Main.kt` has been updated to show a tree.

### Tree Element Definition
For better usability, we decided to split the element into nodes and leaves, obv a leaf cannot have children.

```kotlin
sealed interface Element<T> {

    val data: T
    val depth: Int
    val parent: Element<T>?
    val childIndex: Int
    var next: Element<T>?
    var previous: Element<T>?
    val id: Any
}

class Leaf<T>(
    overrides..
) : Element<T>

class Node<T>(
    overrides...
    private val childrenGenerator: (parent: Node<T>) -> List<Element<T>>,
) : Element<T> 
```

### Tree APIs for ID
Commonly we expect an ID of type `Any` for elements, but nothing binds an element to a tree, so we decide to create an ID strategy where the function `idPath()` can identify an element inside a specific `Tree` (useful to support persistence interactions, if needed)

```kotlin
fun path() = buildList {
    var next: Element<T>? = this@Element
    while (next != null) {
        add(next)
        next = next.parent
    }
}.reversed()

fun idPath() = path().map { it.id }
```

As we expect Tree has no cognition about it is represented, this duty will be done by the `TreeState`, in detail in our implementation we decided to represent the Tree as an ordered list of elements, where the kinship is represented graphically with the depth of every single element... the pair <position, depth> can explain without any doubt the Tree structure.

### TreeState

```kotlin
class TreeState(val delegate: SelectableLazyListState) : FocusableState by delegate {

    internal var tree: Tree<*> by Delegates.notNull()
    fun attachTree(tree: Tree<*>) {
        this.tree = tree
        refreshFlattenTree()
    }

    var flattenedTree by mutableStateOf<List<Tree.Element<*>>>(emptyList())  //this is the state to represent!

    val isFocused -> delegate.isFocused
    val selectedElementsMap -> delegate.selectedItemIndexes

    fun openCloseNodes()
}
```
The TreeState defines all APIs that are strictly bonded to a Tree interaction and delegates the other generic action to the subComponentState like selection->SelectableState and focus->FocusState

### SubComponent: SelectableLazyList
It is made to hold and manage the concept of selection, every Item in the list should be provided with an id (ANY) and the specification that is Selectable or NotSelectable.
Note the focus interaction is Delegate to FocusableLazyColumnState

```kotlin
sealed interface SelectableKey {

    val key: Any
    @JvmInline value class Selectable(override val key: Any) : SelectableKey
    @JvmInline value class NotSelectable(override val key: Any) : SelectableKey
}
```

This indication will be used in the SelectableLazyListState for all selection interactions like mouse/tap events and keybinding events

```kotlin
class SelectableLazyListState(
    internal val delegate: FocusableLazyListState = FocusableLazyListState(),
    val isMultiSelectionAllowed: Boolean = true
) : FocusableState by delegate {

    internal val selectedIdsMap = mutableStateMapOf<SelectableKey, Int>()
    internal var keys = emptyList<SelectableKey>()

    //edit selection apis
    fun selectNext()....
}
```

The binding/refresh from state and list elements will be done by the component via `attachKeys()`:

```kotlin
internal fun attachKeys(keys: List<SelectableKey>, uiId: String) {
      if (this.uiId == null) {
          this.uiId = uiId
      } else {
          require(this.uiId == uiId) {
              "Do not attach the same ${this::class.simpleName} to different SelectableLazyColumns."
          }
      }
      this.keys = keys
      keys.forEachIndexed { index, key ->
          selectedIdsMap.computeIfPresent(key) { _, _ -> index }
      }
  }
```

The mapping of the items will be done with the same pattern as `LazyColumn`, with `SelectableLazyListScope` that forward this mapping downside to `FocusableLazyColumn`:

```kotlin
interface SelectableLazyListScope {

    fun item(
        key: Any,
        contentType: Any? = null,
        focusable: Boolean = true,
        selectable: Boolean = true,
        content: @Composable SelectableLazyItemScope.() -> Unit
    )

    fun items(
        count: Int,
        key: (index: Int) -> Any,
        contentType: (index: Int) -> Any? = { null },
        focusable: (index: Int) -> Boolean = { true },
        selectable: (index: Int) -> Boolean = { true },
        itemContent: @Composable SelectableLazyItemScope.(index: Int) -> Unit
    )

    fun stickyHeader(
        key: Any,
        contentType: Any? = null,
        focusable: Boolean = false,
        selectable: Boolean = false,
        content: @Composable SelectableLazyItemScope.() -> Unit
    )
}
```

**SubComponent: FocusableLazyColumn**
The last layer between Selection and the LazyColumn is made by FocusableLazyList and his state holder FocusableLazyListState, 

```
class FocusableLazyListState(val delegate: LazyListState = LazyListState()) : FocusableState, ScrollableState by delegate {
    internal val lastFocusedKeyState: MutableState<LastFocusedKeyContainer> = mutableStateOf(LastFocusedKeyContainer.NotSet)
    internal val lastFocusedIndexState: MutableState<Int?> = mutableStateOf(null)


    override suspend fun focusItem(itemIndex: Int, animateScroll: Boolean, scrollOffset: Int) {
        val visibleRange = visibleItemsRange.drop(2).dropLast(4)

        if (itemIndex !in visibleRange && visibleRange.isNotEmpty()) {
            when {
                itemIndex < visibleRange.first() -> delegate.scrollToItem(max(0, itemIndex - 2), animateScroll, scrollOffset)
                itemIndex > visibleRange.last() -> {
                    val indexOfFirstVisibleElement = itemIndex - visibleRange.size
                    delegate.scrollToItem(
                        min(delegate.layoutInfo.totalItemsCount - 1, indexOfFirstVisibleElement - 1),
                        animateScroll,
                        scrollOffset
                    )
                }
            }
        }

        focusVisibleItem(itemIndex)
    }
}
```
Like selectableLazyColumn, we are mapping the item that will be forwarded to LazyColumnScope, for every item a FocusRequester will be associated once!
```
interface FocusableLazyListScope {

    fun item(
        key: Any,
        contentType: Any? = null,
        focusable: Boolean = true,
        content: @Composable FocusableLazyItemScope.() -> Unit
    )

    fun items(
        count: Int,
        key: ((index: Int) -> Any)? = null,
        contentType: (index: Int) -> Any? = { null },
        focusable: (index: Int) -> Boolean = { true },
        itemContent: @Composable FocusableLazyItemScope.(index: Int) -> Unit
    )

    fun stickyHeader(
        key: Any,
        contentType: Any? = null,
        focusable: Boolean = false,
        content: @Composable FocusableLazyItemScope.() -> Unit
    )
}
       
```

> Final Components APIs

```
@Composable
fun FocusableLazyColumn(
    verticalScroll: Boolean = false,
    state: FocusableLazyListState = rememberFocusableLazyListState(),
    contentPadding: PaddingValues = PaddingValues(0.dp),
    reverseLayout: Boolean = false,
    verticalArrangement: Arrangement.Vertical =
        if (!reverseLayout) Arrangement.Top else Arrangement.Bottom,
    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
    flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
    onKeyPressed: KeyEvent.(Int) -> Boolean = { _ -> false },
    content: FocusableLazyListScope.() -> Unit
)
```
```

@Composable 
fun SelectableLazyColumn(
    verticalScroll: Boolean = false,
    state: SelectableLazyListState = rememberSelectableLazyListState(),
    contentPadding: PaddingValues = PaddingValues(0.dp),
    reverseLayout: Boolean = false,
    verticalArrangement: Arrangement.Vertical = if (!reverseLayout) Arrangement.Top else Arrangement.Bottom,
    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
    flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
    keyActions: KeyBindingScopedActions = DefaultSelectableLazyColumnKeyActions(state),
    content: SelectableLazyListScope.() -> Unit
) {
    LaunchedEffect(keyActions) {
        state.attachKeybindings(keyActions)
    }

    val container = remember(content) { SelectableLazyListScopeDelegate(state).apply(content) }
    val uiId = remember { UUID.randomUUID().toString() }
    LaunchedEffect(container) { state.attachKeys(container.keys, uiId) }
    FocusableLazyColumn(...args..)
}
```
```

@Composable
fun <T> TreeView(
    tree: Tree<T>,
    treeState: TreeState = rememberTreeState(),
    onElementClick: (Tree.Element<T>) -> Unit = { Log.d("click") },
    onElementDoubleClick: (Tree.Element<T>) -> Unit = { Log.d("double click") },
    keyActions: KeyBindingScopedActions = DefaultTreeViewKeyActions(treeState),
    selectionFocusedBackgroundColor: Color,
    selectionBackgroundColor: Color,
    platformDoubleClickDelay: Long = 500L,
    deepIndentDP: Int = 20,
    arrowContent: @Composable (isOpen: Boolean) -> Unit,
    elementContent: @Composable (Tree.Element<T>) -> Unit
) {
    LaunchedEffect(tree) {
        treeState.attachTree(tree)
    }

    val flattenTree = treeState.flattenedTree

    val scope = rememberCoroutineScope()

    val isTreeFocused = treeState.delegate.lastFocusedIndex != null
    val selectionColor by animateColorAsState(if (isTreeFocused) selectionFocusedBackgroundColor else selectionBackgroundColor)
    SelectableLazyColumn(...args...)
}
```


foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/SelectableColumnOnKeyEvent.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.jewel.foundation.lazy.SelectableColumnKeybindings
+
+interface SelectableColumnOnKeyEvent {
+
+    val keybindings: SelectableColumnKeybindings
+
+    /**
+     * Select First Node
+     */
+    suspend fun onSelectFirstItem(currentIndex: Int)
```
ralstondasilva, line 12: I think we don't need currentIndex here

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableLazyColumn.kt
```diff
@@ -0,0 +1,309 @@
...
+                            state.lastFocusedKeyState.value = LastFocusedKeyContainer.Set(entry.key?.invoke(itemIndex))
+                            state.lastFocusedIndexState.value = itemFinalIndex
+                        }
+                    }
+                    .focusable()
+                    .pointerInput(Unit) {
+                        awaitPointerEventScope {
+                            while (true) {
+                                awaitFirstDown(false)
+                                requesters[itemIndex]!!.requestFocus()
```
ralstondasilva, line 238: We could also save the hasFocus state on line 229, and if this item has focus, we don't need to requestFocus. That way if we have an item with multiple focusable children, and if one of them is focused, we won't change the focus.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableLazyColumn.kt
```diff
@@ -0,0 +1,309 @@
...
+                        }
+                    }
+            ) {
+                entry.itemContent(FocusableLazyItemScope(), itemIndex)
+            }
+        } else {
+            entry.itemContent(FocusableLazyItemScope(), itemIndex)
+        }
+        if (state.lastFocusedIndex == itemFinalIndex) {
+            SideEffect { requesters[itemIndex]!!.requestFocus() }
```
ralstondasilva, line 250: Maybe this could be a LaunchedEffect instead of a SideEffect. 

We could also add the hasFocus check here, and only requestFocus if this item does not have focus.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -0,0 +1,465 @@
...
+     * [InteractionSource] that will be used to dispatch drag events when
+     * this list is being dragged. If you want to know whether the fling (or
+     * animated scroll) is in progress, use [isScrollInProgress].
+     */
+    val interactionSource
+        get() = delegate.interactionSource
+
+    var keybindings: SelectableColumnKeybindings = DefaultSelectableColumnKeybindings
+
+    fun attachKeybindings(keybindings: KeyBindingScopedActions) {
```
andkulikov, line 111: do you want this function to be internal so others can't call it manually?

andkulikov (commented):
Hey! I am the owner of LazyColumn in Google. Jim asked to take a look on this commit so I left a few comments

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableLazyColumn.kt
```diff
@@ -0,0 +1,316 @@
...
+        if (!reverseLayout) Arrangement.Top else Arrangement.Bottom,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
+    flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    onKeyPressed: KeyEvent.(Int) -> Boolean = { _ -> false },
+    content: List<Entry>
+) {
+    Box(
+        modifier
+            .onPreviewKeyEvent { event -> state.lastFocusedIndex?.let { onKeyPressed(event, it) } ?: false }
```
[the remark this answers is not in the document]
lamba92: It was originally recommended by @ralstondasilva, I do not recall exactly why. Of course we can

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableListState.kt
```diff
@@ -0,0 +1,298 @@
...
+}
+
+private suspend fun LazyListState.scrollToItem(index: Int, animate: Boolean, scrollOffset: Int = 0) =
+    if (animate) animateScrollToItem(index, scrollOffset) else scrollToItem(index, scrollOffset)
+
+sealed interface FocusableKey {
+
+    val key: Any?
+
+    class Focusable(val focusRequester: FocusRequester, override val key: Any?) : FocusableKey
```
andkulikov, line 198: this pattern worried me a bit as well
keys and LazyListItemInfo were not intended to contain anything you can perform an action on. it is just a read only info about the last layout pass. but here we allow calling requestFocus() on an object we got from the infos. It feels ok to have something like FocusableLazyListState.focusItem(), but not expose FocusRequesters directly in such manner. maybe this part can be hidden from the public api somehow?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableLazyColumn.kt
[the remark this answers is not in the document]
ralstondasilva: We might be able to avoid FocusableLazyColumn and add this logic to SelectableLazyColumn. Can you help us understand the main use-case for FocusableLazyColumn?

Is it that we want to first focus on the column and then move focus among the contents? or because we want to focus on the column before the contents are loaded or perhaps we want to move focus to the column when the focused item is disposed?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
fscarponi: This was a forgetfulness! will be internal

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/FocusableLazyColumn.kt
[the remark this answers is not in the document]
fscarponi: After all, can be better to fuse Selectable and Focusable, with less code easiest to maintain. I will start with this!

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/VerticalScrollerForLazyColumn.kt
```diff
@@ -0,0 +1,194 @@
...
+        val contentHeight = contentPlaceable.maxByOrNull { it.height }?.height ?: 0
+        val contentWidth = contentPlaceable.maxByOrNull { it.width }?.width ?: 0
+
+        val scrollerConstraints = constraints.copy(minWidth = 0, minHeight = 0)
+        val scrollerPlaceable = subcompose("scroller") {
+            val layoutInfo = listState.layoutInfo
+            val showScroller = layoutInfo.visibleItemsInfo.size < layoutInfo.totalItemsCount
+            if (!showScroller) return@subcompose
+
+            val thumbTopPadding = with(LocalDensity.current) { topContentPadding.toPx() }
```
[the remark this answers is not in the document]
fscarponi: Right! thanks, we discussed about this and we are going to drop the custom scrollbar in favor of the standard one that should be easy to bridge in swing

fscarponi:
The model and its implementation have been revised following the received feedback.

in details:

1. FocusableLazyColumn has been removed, his logic has been moved to Selectable Lazy column.
2. The SelectableLazyListScope will provide to content composition "isFocused" and "isSelected"
3. The new implementation should allow us to execute the user-provided content lambda directly inside the content lambda passed to LazyColumn
4. The KeyBinding and PointerEvent interaction has been refactored in order to be more user-friendly

rock3r:
@andkulikov @ralstondasilva could you take another look to see if this is good to go?

rock3r:
@andkulikov @ralstondasilva ping, this is blocking folks at JetBrains 🙏

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -0,0 +1,226 @@
...
+    private fun Modifier.focusable(key: SelectableKey.Focusable, isFocused: Boolean) =
+        focusRequester(key.focusRequester)
+            .onFocusChanged {
+                if (it.hasFocus) {
+                    state.lastFocusedKeyState.value = SelectableLazyListState.LastFocusedKeyContainer.Set(key)
+                    state.lastFocusedIndexState.value = state.keys.indexOf(key)
+                }
+            }
+            .focusable()
+            .pointerInput(Unit) {
```
andkulikov, line 97: if you want the coroutine block to restart when isFocused changes you should add it as a param for pointerInput. or  use rememberUpdatedState. see https://developer.android.com/reference/kotlin/androidx/compose/ui/input/pointer/package-summary#(androidx.compose.ui.Modifier).pointerInput(kotlin.Any,kotlin.Any,kotlin.coroutines.SuspendFunction1)

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListState.kt
```diff
@@ -0,0 +1,393 @@
...
+    /**
+     * Focuses on the item at the specified index within the lazy list.
+     *
+     * @param itemIndex The index of the item to focus on.
+     * @param animateScroll Whether to animate the scroll to the focused item.
+     * @param scrollOffset The scroll offset for the focused item.
+     * @param skipScroll Whether to skip the scroll to the focused item.
+     */
+    suspend fun focusItem(itemIndex: Int, animateScroll: Boolean = false, scrollOffset: Int = 0, skipScroll: Boolean = false) {
+        val visibleRange = visibleItemsRange.drop(2).dropLast(4)
```
andkulikov, line 93: this looks a bit strange. if you wanted to only include items which are fully visible there are better options as we know the offsets and min max limits

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -0,0 +1,226 @@
...
+            if (focusable(it)) {
+                SelectableKey.Focusable(FocusRequester(), key(it), selectable(it))
+            } else {
+                SelectableKey.NotFocusable(
+                    key(it),
+                    selectable(it)
+                )
+            }
+        }
+        state.attachKeys(selectableKeys)
```
andkulikov, line 176: I am curious why did you decide to manually set the keys into a state instead of getting the key from the visible item via layout info? with that we can delegate all the storage to the lazy list

andkulikov (commented):
Thanks, left a few comments

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
fscarponi: Ok, isFocused will be used as key!

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListState.kt
lamba92: This is to try to mimic the IntelliJ tree scrolling behaviour. The IJ tree will always leave a pad of 2 items from top or bottom from the currently focused item. 

It is a naive and early attempt to mimic do the same. Could you please how to scroll an item in a place such that it is either at the 3rd spot from the top or from the bottom using offsets? Just use the IntelliJ project tree with your keyboard to understand what we are trying to achive

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
lamba92: We need to have access to the keys of items that are not visible, but they are `private`. If you would modify them so that they are always accessible, that would drastically simplify our work!

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -0,0 +1,115 @@
...
+    verticalArrangement: Arrangement.Vertical = if (!reverseLayout) Arrangement.Top else Arrangement.Bottom,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
+    flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    keyActions: KeyBindingScopedActions = DefaultSelectableLazyColumnKeyActions(state),
+    pointerHandlingScopedActions: PointerEventScopedActions = DefaultSelectableLazyColumnPointerEventAction(state),
+    content: SelectableLazyListScope.() -> Unit
+) {
+    LaunchedEffect(keyActions) {
+        state.attachKeybindings(keyActions)
```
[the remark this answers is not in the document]
fscarponi: Keybindings can be changed by the user, and the recomposition will do the binding job for us... A part of that we will migrate this on disposable effect according to your previous hint

rock3r (approved):
:shipit: we can always come back for further changes later if we get more feedback but we can at least unblock this

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListState.kt
andkulikov: If it is exactly what you wanted to achieve then its ok

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
andkulikov: what do you mean by "not visible", if the item is not composed then the requester is not attached to hierarchy so there is no point to use it. is it for rare use cases when the item is composed, but not in the visibleInfo list? 
this happens for focus search when you focus an item which is out of the bounds, in this case bringIntoView scrolling mechanism will make it visible soon, and the other case is prefetching, we do not consider the prefetched item to be focusable as it is not yet placed

PR #46, opened 2023-06-01
New implementation by defaults and colors

devkanro:


core/src/main/kotlin/org/jetbrains/jewel/internal/Stroke.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.isUnspecified
+import androidx.compose.ui.unit.Dp
+
+sealed class Stroke {
+    @Immutable
+    object None : Stroke() {
+
+        override fun toString(): String {
+            return "Stroke()"
```
rock3r, line 15: Nit: I wouldn't override these `toString()`s since we're missing the actual type. I get that was the intent, but I don't think it's a good idea in this case

core/src/main/kotlin/org/jetbrains/jewel/internal/IntelliJColors.kt
```diff
@@ -0,0 +1,92 @@
...
+            panelArea = panelArea,
+            contentArea = contentArea,
+            darkArea = darkArea,
+            disableArea = disableArea,
+            separator = separator,
+            isLight = isLight
+        )
+    }
+
+    override fun toString(): String {
```
rock3r, line 56: Do we need the override? It looks identical to the stock `toString()`

core/src/main/kotlin/org/jetbrains/jewel/internal/Stroke.kt
```diff
@@ -0,0 +1,68 @@
...
+}
+
+fun Stroke(width: Dp, color: Color, alignment: Stroke.Alignment): Stroke {
+    if (width.value == 0f) return Stroke.None
+    if (color.isUnspecified) return Stroke.None
+
+    return Stroke.Solid(width, color, alignment)
+}
+
+fun Stroke(width: Dp, brush: androidx.compose.ui.graphics.Brush, alignment: Stroke.Alignment): Stroke {
```
rock3r, line 55: Import `Brush`

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/Button.kt
```diff
@@ -0,0 +1,94 @@
...
+    @Composable
+    override fun minHeight(): Dp {
+        return MinHeight
+    }
+
+    @Composable
+    override fun buttonColors(): ButtonColors {
+        val palette = LocalIntUiPalettes.current
+        return remember(palette) {
+            org.jetbrains.jewel.internal.buttonColors(
```
rock3r, line 59: Import `buttonColors`

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/Button.kt
```diff
@@ -0,0 +1,94 @@
...
+                ),
+                focusedStroke = Stroke(1.dp, palette.grey().mainBackground(), Stroke.Alignment.Outside),
+                focusHoloStroke = Stroke(3.dp, palette.blue().buttonDefault(), Stroke.Alignment.Outside),
+            )
+        }
+    }
+
+    @Composable
+    override fun outlineButtonColors(): ButtonColors {
+        TODO("Not yet implemented")
```
rock3r, line 92: Let's not forget to add it — although I am not sure what an outline button is. Is it a "normal" button, called "secondary" in the Int UI specs, whereas the `buttonColors()` above represent the "default" button, aka "primary"?

If that's the case, I would have the "primary" button be the special case, since there can only be one per screen, and the "secondary" button be the default for a button. This also reflects how they're implemented in Swing.

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/DarkPalette.kt
```diff
@@ -0,0 +1,274 @@
+package org.jetbrains.jewel.themes.intui.standalone.internal
+
+import androidx.compose.ui.graphics.Color
+import org.jetbrains.jewel.internal.IntelliJColors
+
+class DarkPalette(
```
rock3r, line 6: Here and in the `LightPalette`, instead of hardcoding the values, I'd instead refer to a colour palette with the same colour names you find in Figma and in the actual Swing themes. E.g., `IntUiColorPalette` that contains `Light` and `Dark`, and each have `Grey1` `Grey2`, etc etc

It's trivial to generate them from [`expUI_dark.theme.json`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-resources/src/themes/expUI/expUI_dark.theme.json) and [`expUI_light.theme.json`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-resources/src/themes/expUI/expUI_light.theme.json), and they keep these palettes much tidier, avoiding copy-paste errors and making it easier to then go and add more entries here.

rock3r (changes requested):
Great job! I have mostly minor comments. The only two changes I'd make are on the Button colours and the palettes.

core/src/main/kotlin/org/jetbrains/jewel/internal/Border.kt
```diff
@@ -0,0 +1,373 @@
+package org.jetbrains.jewel.internal
```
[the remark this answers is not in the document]
devkanro: Currently I just write the new implementation in this package so that @fscarponi can continue his work, after finishing most of the controls, I will migrate it out and replace the old controls

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/Button.kt
devkanro: OK, at present, I am referring to the code of material design, and theirs is to use the primary button as the default button. But I think what you said also makes sense.

rock3r: True, but I think this is something we should be coherent with the Darcula/IntUI implementation in Swing.

Material is a great source of inspiration in how to implement the theme when it comes to code structure and general patterns, and we should indeed follow that same approach; where it comes to higher level concepts though, we have more freedom. By choosing consistency with the Swing world we not only make it easier for folks to transition to Jewel, but also make our lives easier on the bridge side, as we don't have to think in reverse :)

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/DarkPalette.kt
rock3r: To be clear, I don't think we should do it at runtime, but it should be trivially easy to parse that JSON and generate the corresponding Kotlin. For now, it can be done by hand in less than 1h. In the future we may want to automate it, but it's not necessary yet I reckon

core/src/main/kotlin/org/jetbrains/jewel/internal/Stroke.kt
devkanro: These toString methods help us record these values in the modifier's `debugInspectorInfo`

rock3r: Sure, but the `object`'s one would produce `None` — which is more useful than `Stroke()` as it contains the actual type and is less confusing — and the auto-generated ones for data classes would look identical, but also include the correct type instead of the generic `Stroke`.

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/internal/DarkPalette.kt
devkanro: OK, a gradle build script can do this

core/src/main/kotlin/org/jetbrains/jewel/internal/IntelliJColors.kt
devkanro: You are right, kotlin data class provides this implementation automatically, I would remove these unnecessary toString

fscarponi (approved):

rock3r (approved):
Sorry for the delay, been busy 😓

PR #69, opened 2023-06-21
ProrgessBar Implementation

fscarponi:
https://github.com/JetBrains/jewel/assets/36624359/331e5edb-d51d-47e4-9b38-0c192167c081

https://github.com/JetBrains/jewel/assets/36624359/29503837-d783-44c1-bf8f-c2a3ff414860


core/src/main/kotlin/org/jetbrains/jewel/LinearProgressBar.kt
```diff
@@ -0,0 +1,215 @@
...
+@Composable
+fun AnimatedDefiniteLinearProgressBar(
+    modifier: Modifier = Modifier,
+    targetProgress: Float, // from 0 to 1
+    defaults: ProgressBarDefaults = IntelliJTheme.progressBarDefaults,
+    colors: ProgressBarColors = defaults.colors()
+) {
+    val progress by animateFloatAsState(
+        targetValue = targetProgress,
+        animationSpec = tween(durationMillis = 250, easing = LinearEasing)
```
rock3r, line 40: I would not use a Linear easing in this case, but rather leave the default, as it looks much better when you have actual easing in/out (for the indeterminate case, it's right to use linear because we want the velocity to be constant)

core/src/main/kotlin/org/jetbrains/jewel/LinearProgressBar.kt
```diff
@@ -0,0 +1,215 @@
...
+import androidx.compose.ui.graphics.TileMode
+import androidx.compose.ui.graphics.drawOutline
+import androidx.compose.ui.unit.Dp
+
+@Composable
+fun AnimatedDefiniteLinearProgressBar(
+    modifier: Modifier = Modifier,
+    targetProgress: Float, // from 0 to 1
+    defaults: ProgressBarDefaults = IntelliJTheme.progressBarDefaults,
+    colors: ProgressBarColors = defaults.colors()
```
rock3r, line 36: If you're already passing in the defaults, you shouldn't need to also pass in the colours. I'd remove this from all composables.

core/src/main/kotlin/org/jetbrains/jewel/LinearProgressBar.kt
```diff
@@ -0,0 +1,215 @@
...
+    trackColor: Color,
+    brushColorList: List<Color>
+) {
+    val shape = defaults.clipShape()
+    val height = defaults.height()
+    Box(
+        modifier
+            .height(height)
+            .clip(shape)
+            .drawWithContent {
```
rock3r, line 81: This is pretty cheap, but in general, we could consider drawWithCache in more complex scenarios (or when we do perf optimisations)

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindow.kt
```diff
@@ -156,6 +161,15 @@ internal class JewelDemoToolWindow : ToolWindowFactory, DumbAware {
                             checked = checked,
                             onCheckedChange = { checked = it }
                         )
+
+                        // TODO progressbar compose
```
rock3r, line 165: You're missing the Compose one here :)

devkanro (approved):

devkanro:
Nicely done

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -81,18 +76,10 @@ fun main() = singleWindowApplication(
                 RadioButtons()
                 Links()
                 TextAreas()
+                ProgressBar()
                 TextFields()
+                ChipsAndTree()
                 // take this at the end, because it's a bit taller
```
rock3r, line 82: We should remove this comment now that it's attached to nothing

rock3r:
I was reviewing this :( @fscarponi please address the comments in a separate PR

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/light/LightProgressBarDefaults.kt
```diff
@@ -0,0 +1,26 @@
+@file:Suppress("MagicNumber")
```
rock3r, line 1: We should really suppress that Detekt rule... I always forget

rock3r (changes requested):

core/src/main/kotlin/org/jetbrains/jewel/LinearProgressBar.kt
fscarponi: Sorry but i'm not convinced that removing it is the best option. I understand that seems to be redundant, but in your idea, we will sacrifice usability for good looking api.

Let's watch a common scenario: a user/dev is creating his plugin, and he wants just one different color for the progress bar (for a specific purpose not related to an ide loading operation)
Those are the 2 patterns in comparison, the first does not expose colors parameter, the second one does
image

rock3r: Is there a specific reason why the defaults aren't a data class? I also agree you shouldn't have to implement a bloody object. But in an ideal scenario, you'd only need to do `ProgressBarDefaults(colours = ...)`, or `whateverDefaultsFromTheTheme.copy(colours = ...)`

rock3r: This is to say, the problem to me seems that the defaults are implemented with an hostile pattern, and exposing the colours is a patch for the symptom rather than a fix for the problem

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindow.kt
fscarponi: something is broken in this file, there is no way to let the ide start the analysis, so it is like coding with eyes closed. i will just put the invocation there for now

rock3r: It happens sometimes and it's super annoying! Usually invalidating caches helps for me

core/src/main/kotlin/org/jetbrains/jewel/LinearProgressBar.kt
fscarponi: defaults are functions that usually take a state as a parameter.

ex
fun backGroudColor(state: ComponentState)

when ComponentState is dataclass like this

data class ComponentState(
val isFocused,
val isSelected,
val isHovered...)

so in the background function, you can simply run a when on the state....

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindow.kt
fscarponi: it probably is something else, when i cancel the cache, it starts analyzing, and an exception is thrown by the IDE...

rock3r: Sigh :(

themes/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/themes/intui/standalone/light/LightProgressBarDefaults.kt
rock3r: Filed in #72

PR #75, opened 2023-06-28
Integration fixes

fscarponi:
Starting to integrate Jewel into the IDE, I encountered some issues that need to be addressed.
1 - In the IDE, managing the focus of SelectableLazyColumn causes an exception that we believed was resolved with an uninitialized FocusRequester. Since it is a non-critical operation, it has been caught in a catch block.
2 - The ResourceLoader.Default system was failing to load resources in the IDE. Therefore, thanks to @lamba92  we created and made available the ability to customize the ResourceLoader by providing some convenience functions
3 - Minor fixes.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/jewel/Icon.kt
+++ b/core/src/main/kotlin/org/jetbrains/jewel/Icon.kt
@@ -32,6 +32,7 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.unit.dp
 import org.xml.sax.InputSource
 import java.io.InputStream
+import java.util.jar.JarFile
 
 /**
  * Icon component that draws [imageVector] using [tint], defaulting to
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/jewel/Icon.kt
```diff
@@ -218,3 +219,30 @@ private fun Size.isInfinite() = width.isInfinite() && height.isInfinite()
...
+        // Class not from a JAR
+        return null
+    }
+    return classPath.substringBefore("!").removePrefix("jar:file:")
+}
+
+fun extractFileFromJar(jarPath: String, filePath: String) =
+    JarFile(jarPath).use { jar ->
+        jar.getEntry(filePath)?.let { entry ->
+            jar.getInputStream(entry).use { it.readBytes().inputStream() }
```
rock3r, line 236: I'm not sure why you're reading all bytes from an input stream just to create an input stream from it?

core/src/main/kotlin/org/jetbrains/jewel/Icon.kt
```diff
@@ -218,3 +219,30 @@ private fun Size.isInfinite() = width.isInfinite() && height.isInfinite()
 
 // Default icon size, for icons with no intrinsic size information
 private val DefaultIconSizeModifier = Modifier.size(16.dp)
+
+fun getJarPath(klass: Class<*>): String? {
```
rock3r, line 223: Do all these new things need to be public? I don't think we need them in our API

PR #87, opened 2023-07-17
Patch SVG files

rock3r:
* Patch colours in SVGs while loading them
* Logic to load the right icon file based on state and theme (1st pass)
* Import the Intellij Light (old UI) SVG files in standalone Int UI

Sorry for the noise, not sure how it happened, I was sure the style changes were already on main, but apparently I was wrong.

Main files to review: `IntelliJSvgPatcher`, `ResourcePathPainterProvider`, `PaletteMapperFactory`, and `PaletteMapper`. The rest is boring stuff that comes off of those :)

PR #89, opened 2023-07-18
Cache icon painters

rock3r:
We now have an extremely simple, but very functional, cache for the painters generated by patching the SVGs. This is the same thing the IDE does, at a very high level — but with the difference that they mostly cache rasterised icons, and they also pay MUCH more attention to the origin classloader. This code assumes that a given icon at a given path is only going to be present on a given classloader, which seems like a reasonable assumption for now. With us storing SVGs instead of rasterised images, too, we don't really need to go back and re-read the SVG if we later wanted to store rasterised renditions instead.

We may need to make this mechanism smarter in the future, but it's fine for now.

This PR also introduces some related refactorings, tying up the painter provider implementations that generate from resources. This way we don't have several slightly different ways to do it anymore, but just the one way.


rock3r:
A note: [0ee5115](https://github.com/JetBrains/jewel/pull/89/commits/0ee51157aebfc72e6f9e11bcf300fef6e74f71cd) introduces a fix for an issue identified in #87 caused by the wrong order of modifiers fallback in image names.

fscarponi (approved):

Issue #79, opened 2023-06-30
Tab & TabStrip


rock3r:
* Tab component
* Basic scrollable TabStrip component

Note: TabStrip doesn't need to handle content changes. It's limited to the tab strip only.

![screenshot of IntUI tab specs](https://github.com/JetBrains/jewel/assets/153802/4afaf3c5-cc7b-4916-9b47-be29f34cf6a5)

fscarponi:
<img width="53" alt="image" src="https://github.com/JetBrains/jewel/assets/36624359/aac4cfbe-9fe8-4750-9587-a5a9ecbf3910">
 it is rounded in the IDE

rock3r:
I swear I had _never_ noticed it is 😱

rock3r:
Moving to m3 since #54 just landed and @fscarponi doesn't have time to do this before the cutoff

PR #90, opened 2023-07-31
Fixes on Tree Chip and TextColor Handling, impl for Tabs and TabStrip

fscarponi:
```kotlin
TabStrip(Modifier.fillMaxWidth()) {
            tab(
                tabNameString = "Tab 1",
                closable = false,
                selected = selectedTab == 0,
                tabIconResource = "icons/kotlin.svg",
                onTabClick = { selectedTab = 0 }
            )
            tabs(
                10,
                selected = { selectedTab == it + 1 },
                closable = false,
                tabNameString = { "tab " + (it + 1) },
                tabIconResource = { "icons/kotlin.svg" },
                onTabClick = { selectedTab = it + 1 },
                onTabClose = { },
                styleType = { if (it) TabStyleType.Default else TabStyleType.Editor }
            )
           editorTab(
                tabNameString = "Tab 1",
                closable = false,
                selected = selectedTab == 0,
                tabIconResource = "icons/kotlin.svg",
                onTabClick = { selectedTab = 0 }
            )
            ditorTabs(
                10,
                selected = { selectedTab == it + 1 },
                closable = false,
                tabNameString = { "tab " + (it + 1) },
                tabIconResource = { "icons/kotlin.svg" },
                onTabClick = { selectedTab = it + 1 },
                onTabClose = { },
                styleType = { if (it) TabStyleType.Default else TabStyleType.Editor }
            )
        }
```


core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,180 @@
...
+                        closable(it),
+                        styleType(it),
+                        { onTabClose(it) },
+                        { onTabClick(it) }
+                    )
+                }
+            }
+        }
+    )
+}
```
rock3r, line 180: I'd recommend a simpler approach, like the one done by lazy lists: create an internal/private `TabStripScopeImpl` class that implements `TabStripScope` and stores the items added in a private backing mutable list. You could even use the same `[Mutable]IntervalList<T>` APIs that the lazy lists use, but you don't necessarily need to. A simple mutable list would work fine.

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,179 @@
...
+    tabIconResource = tabIconResource,
+    closable = closable,
+    interactionSource = interactionSource,
+    tabStyle = tabStyle,
+    onTabClose = onTabClose,
+    onTabClick = onTabClick
+)
+
+@Composable
+internal fun TabImpl(
```
rock3r, line 76: Should be private

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,179 @@
...
+    tabNameString: String,
+    tabIconResource: String?,
+    closable: Boolean,
+    interactionSource: MutableInteractionSource,
+    tabStyle: TabStyle,
+    onTabClose: () -> Unit,
+    onTabClick: () -> Unit,
+) {
+    var tabState by remember { mutableStateOf(TabState.of(selected = selected)) }
+    remember(selected) { tabState = tabState.copy(selected = selected) }
```
rock3r, line 87: This looks extra dodgy — why do we need remember on something that isn't used anywhere? Should this be an effect? But even then, this seems to be doing what the previous line is doing anyway

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,179 @@
...
+    val lineColor by tabStyle.colors.underlineTintFor(tabState)
+    val lineThickness = remember { tabStyle.metrics.underlineThickness }
```
rock3r, line 89: Why is this remembered? It's not a state. It only affects drawing as well. I have a vague recollection of this being a workaround for something but I don't remember what :)

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
```diff
@@ -0,0 +1,153 @@
...
+    val closeHoveredTint: Color
+    val closeHoveredFocusedTint: Color
+    val closeHoveredSelectedTint: Color
+    val closeHoveredFocusedSelectedTint: Color
+    val closeFocusedTint: Color
+    val closeFocusedSelectedTint: Color
+    val closeSelectedTint: Color
+
+    @Composable
+    fun backGroundColorFor(state: TabState) = rememberUpdatedState(
```
rock3r, line 93: It's `background`, not `backGround`

core/src/main/kotlin/org/jetbrains/jewel/styling/LazyTreeStyling.kt
```diff
@@ -28,11 +28,26 @@ interface LazyTreeColors {
     val elementBackgroundSelected: Color
     val elementBackgroundSelectedFocused: Color
 
+    val elementForeground: Color
```
rock3r, line 31: We use `content`, not `foreground`

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,179 @@
...
+            .clickable {
+                onTabClick()
+                focusRequester.requestFocus()
+            }
+            .onPointerEvent(PointerEventType.Enter) {
+                tabState = tabState.copy(hovered = true)
+            }
+            .onPointerEvent(PointerEventType.Exit) {
+                tabState = tabState.copy(hovered = false)
+            }
```
rock3r, line 112: You should use `LaunchedEffect(interactionSource) { ... }` to update the state! This is wrong (and expensive). See how other components do this.

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,179 @@
...
+            .drawBehind {
+                val strokeThickness = lineThickness.toPx()
+                val startY = size.height - strokeThickness
+                val endX = size.width
+                val capDxFix = 3f
```
rock3r, line 117: Not sure what this magic number is for... Seems fishy we can take arbitrary thicknesses but this is hardcoded

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
```diff
@@ -0,0 +1,153 @@
...
+@Immutable
+interface TabIcons {
+
+    val close: String
+
+    @Composable
+    fun closePainter(resourceLoader: ResourceLoader) = painterResource(close, resourceLoader)
+}
+
+interface TabMetrics {
```
rock3r, line 32: Missing `@Stable`/`@Immutable`

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    val underlineThickness: Dp
+    val elementContentPadding: PaddingValues
+    val elementMinHeight: Dp
+    val closeContentGap: Dp
+}
+
+interface TabColors {
```
rock3r, line 40: Missing `@Stable`/`@Immutable`

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
```diff
@@ -0,0 +1,153 @@
...
+            state.isHovered -> tabUnderlineHovered
+            state.isSelected -> tabUnderlineSelected
+            else -> tabUnderline
+        }
+    )
+}
+
+@Immutable
+@JvmInline
+value class TabState(val state: ULong) {
```
rock3r, line 137: This is not tracking a number of actual valid states — see comments above re: this being a `SelectableComponentState`

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
```diff
@@ -0,0 +1,176 @@
...
+    val tabUnderlineSelected: Color
+
+    val closeTint: Color
+    val closeHoveredTint: Color
+    val closeHoveredFocusedTint: Color
+    val closeHoveredSelectedTint: Color
+    val closeHoveredFocusedSelectedTint: Color
+    val closeFocusedTint: Color
+    val closeFocusedSelectedTint: Color
+    val closeSelectedTint: Color
```
rock3r, line 76: This is wrong — you shouldn't be tinting the icon. There are different icons for different states:

image

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/Tabs.kt
```diff
@@ -0,0 +1,57 @@
...
+//                selected = { selectedTab == it + 1 },
+//                closable = { false },
+//                tabNameString = { "tab " + (it + 1) },
+//                tabIconResource = { "icons/kotlin.svg" },
+//                onTabClick = { selectedTab = it + 1 },
+//                onTabClose = { },
+//                styleType = { Random.nextBoolean().let { if (it) TabStyleType.Default else TabStyleType.Editor } }
+//            )
+//        }
+//    }
```
rock3r, line 36: Can we delete this if it's unused?

samples/standalone/src/main/resources/icons/kotlin.svg
rock3r, line 1: Nit: we should likely be using the [current logo](https://kotlinlang.org/docs/kotlin-brand-assets.html#kotlin-logo), not the old one

rock3r (changes requested):

core/src/main/kotlin/org/jetbrains/jewel/Dropdown.kt
```diff
@@ -63,11 +64,11 @@ fun Dropdown(
...
 
         var dropdownState by remember(interactionSource) {
-            mutableStateOf(DropdownState.of(enabled = enabled, error = error))
+            mutableStateOf(DropdownState.of(enabled = enabled, outline = outline))
         }
 
-        remember(enabled, error) {
-            dropdownState = dropdownState.copy(enabled = enabled, error = error)
+        remember(enabled, outline) {
+            dropdownState = dropdownState.copy(enabled = enabled, outline = Outline.Error)
```
rock3r, line 71: Added #91 to track that we need to move the outline out of the component state

core/src/main/kotlin/org/jetbrains/jewel/Menu.kt
```diff
@@ -489,8 +494,9 @@ value class MenuItemState(val state: ULong) : InteractiveComponentState {
         focused: Boolean = isFocused,
         hovered: Boolean = isHovered,
         pressed: Boolean = isPressed,
+        active: Boolean = isActive,
     ): MenuItemState =
-        of(selected, enabled, focused, hovered, pressed)
+        of(selected, enabled, focused, hovered, pressed, active)
 
     override fun toString() =
```
rock3r, line 501: Need to update this too

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,327 @@
...
+                is HoverInteraction.Exit -> tabStripState = tabStripState.copy(hovered = false)
+                is FocusInteraction.Focus -> tabStripState = tabStripState.copy(focused = true, active = true)
+                is FocusInteraction.Unfocus -> tabStripState = tabStripState.copy(focused = false, active = false)
+            }
+        }
+    }
+    val scrollState = rememberScrollState()
+
+    Box(
+        Modifier
```
rock3r, line 53: This modifier should be applied to the `Row`, and the `Box` removed as unnecessary

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,327 @@
...
+            .focusable(true, interactionSource)
+    ) {
+        Row(
+            modifier = modifier
+                .fillMaxWidth()
```
rock3r, line 58: Don't do this — respect the size set by the `modifier` passed in.

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,293 @@
...
+                is PressInteraction.Press -> closeButtonState = closeButtonState.copy(pressed = true)
+                is PressInteraction.Cancel, is PressInteraction.Release ->
+                    closeButtonState =
+                        closeButtonState.copy(pressed = false)
+
+                is HoverInteraction.Enter -> closeButtonState = closeButtonState.copy(hovered = true)
+                is HoverInteraction.Exit -> closeButtonState = closeButtonState.copy(hovered = false)
+            }
+        }
+    }
```
rock3r, line 90: These should live in the scope where the component they refer to is — in the `if (showCloseIcon)` block.

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,293 @@
...
+                            interactionSource = closeActionInteractionSource,
+                            indication = null,
+                            onClick = tabData.onTabClose,
+                            role = Role.Button
+                        ).size(16.dp),
+                        painter = closePainter,
+                        contentDescription = "Close tab ${tabData.tabNameString}"
+                    )
+                } else {
+                    Box(Modifier.size(16.dp)) {}
```
rock3r, line 153: You could use a `Spacer(Modifier.size(16.dp))` instead, since you have no content.

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
```diff
@@ -0,0 +1,293 @@
...
+    @Suppress("ExpressionBodySyntax")
+    @Composable
+    override fun rememberUpdatedInstance(interactionSource: InteractionSource): IndicationInstance {
+        return NoIndicationInstance
+    }
+}
+
+@Immutable
+@JvmInline
+value class TabCloseButtonState(val state: ULong) : InteractiveComponentState {
```
rock3r, line 177: You could literally just use a `ButtonState` instead of creating a new, identical, one

rock3r (changes requested):

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,180 @@
...
+                .horizontalScroll(scrollState)
+                .scrollable(
+                    orientation = Orientation.Vertical,
+                    reverseDirection = ScrollableDefaults.reverseDirection(
+                        LocalLayoutDirection.current,
+                        Orientation.Vertical,
+                        false
+                    ),
+                    enabled = isHovered,
```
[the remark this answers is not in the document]
fscarponi: This is the behavior of Editor TabStrip in the IDE, can explain how to reproduce a different behavior?

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,180 @@
...
+                    state = scrollState,
```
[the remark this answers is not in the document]
fscarponi: This is the behavior of Editor TabStrip in the IDE if you are over the tabStrip in the editor Tab Strip, vertical scroll will produce a horizontal scroll.

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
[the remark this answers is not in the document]
rock3r: I'm sorry but I don't follow — I don't know how I would repro this behaviour you're talking about to begin with. That's why I am saying this seems wrong. I don't understand why the tab strip should not be scrollable unless it's hovered.

The reason I don't understand is that scroll only works when you hover the scrollable component with the pointer (the children are not focusable, so you can't even use the keyboard to scroll). Or am I mistaken?

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,180 @@
...
+                isHovered = true
+            }
+            .onPointerEvent(PointerEventType.Exit) {
+                isHovered = false
+            }
+    ) {
+        Row(
+            modifier = modifier
+                .fillMaxWidth()
+                .horizontalScroll(scrollState)
```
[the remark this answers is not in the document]
rock3r: See [previous comment](https://github.com/JetBrains/jewel/pull/90#discussion_r1281969099)

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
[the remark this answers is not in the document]
rock3r: I don't think your answer is in any way related to my comment

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
fscarponi: outdate, already done

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
fscarponi: [we already discussed about that] (https://github.com/JetBrains/jewel/issues/79)

core/src/main/kotlin/org/jetbrains/jewel/styling/TabStyling.kt
fscarponi: outdated, already soved

core/src/main/kotlin/org/jetbrains/jewel/styling/LazyTreeStyling.kt
```diff
@@ -28,11 +28,26 @@ interface LazyTreeColors {
...
+    val elementForegroundSelected: Color
+    val elementForegroundSelectedFocused: Color
+
     val chevronTint: Color
     val chevronTintSelected: Color
     val chevronTintFocused: Color
     val chevronTintSelectedFocused: Color
 
+    @Composable
+    fun elementForegroundFor(state: TreeElementState) = rememberUpdatedState(
```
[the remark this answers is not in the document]
fscarponi: outdated

core/src/main/kotlin/org/jetbrains/jewel/Tabs.kt
rock3r: I know _why_ there are rounded caps and I know you have to make the line stay inside the actual tab. That is not what I pointed out. What I pointed out is that 3f is a hardcoded value that will only work when the line is exactly 6 dp wide. The cap "extra bit" is always strokeWidth / 2 wide, so hardcoding it to 3 is wrong.

Not only that, this will probably only work for a specific density value, since it's not scaled based on the density, whereas the other values are (see line 111 converting a dp value to pixels, which is a density-aware operation).

core/src/main/kotlin/org/jetbrains/jewel/styling/LazyTreeStyling.kt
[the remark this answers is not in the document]
rock3r: You know you can just mark things as resolved once you address them, right?

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
fscarponi: doing this, the horizontal scroll bar stop working.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/BasicLazyTree.kt
```diff
@@ -202,7 +202,7 @@ value class TreeElementState(val state: ULong) {
 
         private val Focused = 1UL shl 0
```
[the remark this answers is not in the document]
fscarponi: tree should be reimplemented, it will be done in a future pr

core/src/main/kotlin/org/jetbrains/jewel/styling/ResourcePainterProvider.kt
```diff
@@ -47,7 +47,7 @@ class ResourcePainterProvider<T : InteractiveComponentState>(
 
         if (state.isEnabled) {
             when {
-                state.isFocused -> append("Focused")
+                state is SelectableComponentState && state.isFocused -> append("Focused")
```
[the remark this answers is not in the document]
fscarponi: ups!

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+    val scrollState = rememberScrollState()
+
+    Box(
+        Modifier
+            .focusable(true, interactionSource)
+    ) {
+        Row(
+            modifier = modifier
```
rock3r, line 57: FYI, the `modifier` parameter should only ever be applied to the _one_ composable that sits at the root level — see https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-component-api-guidelines.md#modifier-parameter

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
rock3r: That is because you are using the scrollbars wrong — you should only have one top-level `Box` that contains the `Row` and the `HorizontalScrollbar` (see [the Compose Desktop docs](https://github.com/JetBrains/compose-multiplatform/tree/master/tutorials/Desktop_Components#scrollbars) for an example).

If you simplify your setup to look like the sample, it should work fine.

fscarponi: Ok, after a deep reading of the example, simplifying the code to match the perfect use case of example
```
box{
    box{
      ROW{}
    }
    if (should) HorizontalScroller
}
```
it did not solve the problem because the focus handling does not work as well as it should
( tried also to move the focus(interactionSource) in each container )

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
fscarponi: the mistake was done...

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
fscarponi: ok there was a problem with InteractionSource, using a coroutine to handle the hover event, solved the problem.

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
fscarponi: but correct if i'm wrong… the top Box respects the size sets by the modifier passed in... the inside row will fit the box size...

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
```diff
@@ -0,0 +1,180 @@
...
+                    it.onTabClose,
+                    it.onTabClick
+                )
+            }
+        }
+        if (isHovered) {
+            Box(
+                modifier =
+                Modifier
+                    .matchParentSize()
```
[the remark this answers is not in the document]
fscarponi: already solved

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
rock3r: I don't understand what you mean 😅

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
rock3r: The code looked different when I wrote this, but regardless, if the `modifier` passed in doesn't set a specific width, this will make the `Row` take up all available width, which in turn will make the `Box` do the same. What is the reason you have this modifier? What are you trying to achieve, that doesn't work if the `Row` is just allowed to take up as much space as it needs (up to the point where it makes the parent take the maximum width), which is the default behaviour?

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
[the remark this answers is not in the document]
rock3r: Ping on this

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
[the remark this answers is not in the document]
rock3r: I think you're missing my point:

>[by default] scroll only works when you hover the scrollable component with the pointer

Given that scroll does already exactly the same you seem to want it to do, I don't understand why you need to add the "only make it scroll when hovered". Scrollbars also hide when the mouse isn't hovered. It seems like that entire logic is unnecessary. That is what I am saying.

Put it otherwise: what is exactly that is not working if you completely remove this hover-checking logic? Because it seems to me it works exactly as you're describing it anyway (and, I suspect, this logic may actually break things, but I haven't really checked it in action as there is no sample that could highlight it)

PS: don't mark a conversation as resolved unless it _is_ resolved, makes it very hard to find again

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
rock3r: It really feels like there is something in Compose trying to stop you from doing something you're not supposed to, here  😅

core/src/main/kotlin/org/jetbrains/jewel/TabStrip.kt
fscarponi: ok, now i got what you were trying to explain!

rock3r: Sorry I didn't explain it too well before 😅

PR #92, opened 2023-08-04
LazyTree: moving out Tree.Element from LazyTreeState 

fscarponi:


foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/BasicLazyTree.kt
```diff
@@ -216,3 +223,62 @@ value class TreeElementState(val state: ULong) {
...
+    treeState: TreeState,
+    tree: Tree<*>,
+) {
+    Log.d("-----treeRefreshed----")
+    flattenedTree.value = tree.roots.flatMap { flattenTree(it, treeState.openNodes, treeState.allNodes) }
+    flattenedTree.value.forEach {
+        Log.d(it.idPath().toString())
+    }
+//    delegate.updateKeysIndexes()
+    Log.d("----treeRefreshedPrinted----")
```
rock3r, line 238: Can we remove the debug statements and commented code?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/BasicLazyTree.kt
```diff
@@ -216,3 +223,62 @@ value class TreeElementState(val state: ULong) {
         )
     }
 }
+
+private suspend fun refreshFlattenTree(
```
rock3r, line 227: `refreshFlattenedTree()`? It's also very dodgy to pass in a State, I'd pass in the value and output the new one instead, reassigning the new value where the state is held instead.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/TreeState.kt
```diff
@@ -30,114 +26,94 @@ class TreeState(
...
-                }
-            }
-
-            is Tree.Element.Leaf<*> -> {
-                orderedChildren.add(element)
-            }
-        }
-        return orderedChildren
-    }
+//
```
rock3r, line 35: Can we delete commented out code?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/TreeState.kt
```diff
@@ -161,7 +138,13 @@ class TreeState(
         delegate.deselectAll()
     }
 
-    fun isNodeOpen(element: Tree.Element.Node<*>) = element.isOpen
-
-    fun toggleNode(element: Tree.Element.Node<*>) = if (element.isOpen) closeNode(element) else openNode(element)
+    fun toggleNode(nodeId: Any) {
+        Log.d("toggleNode $nodeId")
```
rock3r, line 142: We should remove these logging statements

rock3r (commented):
Just some comments, then good to go

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/tree/TreeState.kt
fscarponi: the log is set to OFF by default, i will remove it later, when the selectableLazyColumn's focus issue is fixed

rock3r (approved):
ayeaye

PR #95, opened 2023-08-22
Refactor theme generator gradle plugin and optimize sarif results

lamba92:
This commit includes the optimization and refactor of the codebase. In the build.gradle files, the 'intUiLight' and 'intUiDark' themes have been registered and their properties have been updated. The Theme Generation plugin has been changed to use project conventions and the color group reading process was simplified. Sarif results are now generated with kotlinx.sarif, which allows fluent API for handling the sarif format. Also, it includes switching from the java-sarif library to kotlin-sarif.

The changes enhance the robustness and reliability of the code. They also make it easier to understand, modify, and maintain the code. And using Kotlin-based libraries optimizes the performance and simplifies future changes in the project.


themes/int-ui/int-ui-core/build.gradle.kts
```diff
@@ -11,17 +11,15 @@ dependencies {
...
-    val ideaVersion = "232.6734"
-
-    create("intUiLight") {
-        themeClassName("org.jetbrains.jewel.themes.intui.core.theme.IntUiLightTheme")
-        themeFile("platform/platform-resources/src/themes/expUI/expUI_light.theme.json")
-        ideaVersion(ideaVersion)
+    register("intUiLight") {
+        themeClassName.set("org.jetbrains.jewel.themes.intui.core.theme.IntUiLightTheme")
+        themeFile.set("platform/platform-resources/src/themes/expUI/expUI_light.theme.json")
+        ideaVersion.set("232.6734")
```
rock3r, line 17: Why no val anymore?

lamba92: To avoid using `afterEvaluate`. Using `Property`s allows to create tasks right away using the `all` and connecting the properties in cascade, assigning them even before the user fills them in. 

The `val/var` approach is not recommended because it forces to use `afterEvaluate`. Check [here](https://blog.gradle.org/simpler-kotlin-dsl-property-assignment).

lamba92: It was already lol

rock3r (approved):

PR #103, opened 2023-08-31
Refactor UI Component Styling in Jewel to Support defaults

lamba92:
Retrieving text styles requires coroutines. This commit aims to allow the theme to start with text defaults and as soon as text styles get retrieved, the theme gets updated.


rock3r:
@lamba92 please rebase on what I just force pushed (we had conflicts against main)

PR #104, opened 2023-08-31
Refactor UI Component Styling in Jewel to Support defaults

lamba92:
Retrieving text styles requires coroutines. This commit aims to allow the theme to start with text defaults and as soon as text styles get retrieved, the theme gets updated.

### DIFF
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
@@ -14,7 +14,6 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
-import com.intellij.ide.ui.LafManager
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.ui.JBColor
 import com.intellij.util.ui.JBValue
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
@@ -14,6 +14,7 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import com.intellij.ide.ui.LafManager
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.ui.JBColor
 import com.intellij.util.ui.JBValue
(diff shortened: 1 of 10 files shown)


ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -189,3 +190,16 @@ internal operator fun TextUnit.plus(delta: Float) =
         isEm -> TextUnit(value + delta, type)
         else -> this
     }
+
+internal fun isLightTheme(): Boolean {
```
rock3r, line 194: No, this is wrong and completely arbitrary. Use `JBColor.isBright` for this. The value is also already provided by `IntelliJThemeDefinition#isDark` and you should use that.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/Svg.kt
rock3r, line 1: I just deleted this file because it wasn't used anywhere FYI

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
```diff
@@ -1,76 +1,73 @@
...
-import org.jetbrains.jewel.SvgLoader
-import org.jetbrains.jewel.themes.PaletteMapperFactory
 import org.jetbrains.jewel.themes.intui.core.IntUiThemeDefinition
-import org.jetbrains.jewel.themes.intui.core.IntelliJSvgPatcher
 
-@Service(Service.Level.APP)
-class SwingBridgeService : Disposable, CoroutineScope {
+@Service(Level.APP)
+class SwingBridgeService @JvmOverloads constructor(
+    private val coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob() + CoroutineName("JewelSwingBridge")),
```
rock3r, line 22: Does this work correctly now on 232+?

rock3r (changes requested):

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/Svg.kt
lamba92: just tested, turns out, the current IJ dependency injection is not smart enough for this use case and will always choose the no-args constructor. Classic IJ!

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
lamba92: just tested, the current IJ dependency injection is not smart enough for this use case and will always choose the no-args constructor. Classic IJ!

rock3r: Also probably worth suggesting a fix to you-know-who :D

rock3r (approved):

PR #110, opened 2023-09-04
Introduce DataProvider support

Walingar:


ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/DataProviderNode.kt
```diff
@@ -0,0 +1,58 @@
...
+ * So, each [DataProviderNode] provides itself and read parent node.
+ * It allows building tree of [DataProviderNode] and traverse it later on.
+ *
+ * @see ModifierLocalModifierNode
+ */
+private val LocalDataProviderNode = modifierLocalOf<DataProviderNode?> {
+    null
+}
+
+internal class DataProviderNode(var dataProvider: (dataId: String) -> Any?) :
```
Walingar, line 22: Not sure that this implementation is the best especially from correctness and thread safety side, but for now it works. I asked folks to review it

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/ProvideData.kt
```diff
@@ -0,0 +1,109 @@
...
+ * Layout composable that create the bridge between [Modifier.provideData] used inside [content] and
+ * [component]. So, IntelliJ's [DataManager] can use [component] as [DataProvider].
+ *
+ * When IntelliJ requests [getData] from [component] Compose will traverse [DataProviderNode] hierarchy
+ * and collect it according [DataProvider] rules (see javadoc).
+ */
+// TODO: choose better naming?
+@Suppress("unused", "FunctionName")
+@Composable
+fun ComponentDataProviderBridge(
```
Walingar, line 27: I guess we can choose better name :)

rock3r: I'm not sure I understand why we offer both a modifier and a composable for this — I thought a modifier would be sufficient? Is that to simplify the process of registering/unregistering the provider?

rock3r (approved):

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/ProvideData.kt
Walingar, line 27: @rock3r single modifier is useless itself, because we need a way to request DataProvider.getData. And this ComponentDataProviderBridge provide this way through JComponent. Also, we can provide LocalDataProvider if it will be needed in Compose itself. But I decided to keep only this api for now.

Walingar: @rock3r single modifier is useless itself, because we need a way to request DataProvider.getData. And this ComponentDataProviderBridge provide this way through JComponent. Also, we can provide LocalDataProvider if it will be needed in Compose itself. But I decided to keep only this api for now.

rock3r (approved):

PR #112, opened 2023-09-07
Fix icon loading in IDE

rock3r:
Loading in the IDE is a mess — between classloaders, fallbacks, mapping etc it needed a lot of work. This PR fixes it by changing how things work internally. The simplified path calculation has been moved to the ResourceLoader level, instead of the SvgLoader level, and an attempt to work around the differences in classloading between the production IDE and the IDE launched via Gradle has been made.

> [!NOTE]
> tl;dr in prod the icon paths are different: they don't have the `DirProvider.dir()` prefix you get when running from Gradle. So what we do is we try to load both with and without that prefix, which should cover both scenarios. In any case, further testing is needed.

The bridge now has a different ResourceLoader than the non-IDE one, and there are more components in the IDE demo (plus some fixes here and there required for this to work).

This PR is in draft mode because **these changes break standalone icon loading**. I will mark as ready when that's solved.


rock3r:
Standalone icon loading is fixed, but the mapping still doesn't work. I'm looking into that.

fscarponi (approved):

PR #117, opened 2023-09-19
Add condition for rendering dropdown icon

lamba92:
Added a conditional check to render the dropdown chevron icon only when the input field is enabled. This prevents the chevron from being displayed and potentially misleading the user when the dropdown is not selectable.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/jewel/Dropdown.kt
+++ b/core/src/main/kotlin/org/jetbrains/jewel/Dropdown.kt
@@ -132,11 +132,13 @@ fun Dropdown(
                     contentAlignment = Alignment.Center,
                 ) {
                     val chevronIcon by style.icons.chevronDown.getPainter(resourceLoader, dropdownState)
-                    Icon(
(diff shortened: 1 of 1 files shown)


fscarponi (changes requested):
If you are going to remove the icon, you should put a box to hold his space.
a better solution would be to use Trasparent/Custom color for disable icon

rock3r:
Dropdowns do have the icon. It's disabled though. 

image

This is how you do it. https://github.com/JetBrains/jewel/blob/6de2fd7d8cd3a80f99c53919ca263088c8ce71f6/core/src/main/kotlin/org/jetbrains/jewel/Link.kt#L309

rock3r (changes requested):
See my note

fscarponi:
this pr is not more needed, because you can achieve the same result by overriding the theme with a color transparent for a disabled state. i'm going to close this!

rock3r:
@fscarponi did you read my comment? Even the intention behind this change is *wrong*

PR #118, opened 2023-09-19
Refine 'addComposeTab' to include closeable option

lamba92:
Introduced 'isCloseable' parameter in 'addComposeTab' function to enable flexibility in controlling behavior of tabs. This option allows programmatically defining whether a tab can be closed or not. Default value set to false to maintain existing function calls.

### DIFF
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/ComposePanel.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/ComposePanel.kt
@@ -7,7 +7,13 @@ import com.intellij.openapi.wm.ToolWindow
 fun ToolWindow.addComposeTab(
     tabDisplayName: String,
     isLockable: Boolean = true,
+    isCloseable: Boolean = false,
     content: @Composable () -> Unit,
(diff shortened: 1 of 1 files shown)


fscarponi (approved):

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/ComposePanel.kt
```diff
@@ -7,7 +7,13 @@ import com.intellij.openapi.wm.ToolWindow
 fun ToolWindow.addComposeTab(
     tabDisplayName: String,
     isLockable: Boolean = true,
+    isCloseable: Boolean = false,
     content: @Composable () -> Unit,
-) = ComposePanel()
-    .apply { setContent(content) }
-    .also { contentManager.addContent(contentManager.factory.createContent(it, tabDisplayName, isLockable)) }
+) {
+    System.setProperty("compose.swing.render.on.graphics", "true")
```
rock3r, line 13: WTF why is this here

rock3r:
Sigh I guess I have to fix this on main now 🙄

Issue #120, opened 2023-09-20
Drop Darcula-standalone plans


rock3r:
While in the IDE we still support Darcula via the Swing bridge, we don't have it in scope for the standalone version.

This means that we can remove a layer of abstraction from the theme — the interfaces in `core` — and convert them to classes (work done alongside #83). This should not only simplify the tweaking of themes, but also make the code leaner. We're not planning to add further themes for now, and even if we did, they would have to either work with the existing drawing logic — and thus with those styling classes — or require a separate implementation.

rock3r:
This was done in #198

PR #121, opened 2023-09-20
Housekeeping

rock3r:
* Remove old cruft (darcula modules, compose-utils)
* Move int-ui to top level (see #120 — no more standalone Darcula)
* Update readme to reflect current state
* Clean up config plugins not to depend on each other


rock3r:
@Walingar any clue why the standalone sample crashes on startup with:

```
org.jetbrains.skiko.LibraryLoadException: Cannot find libskiko-macos-arm64.dylib.sha256, proper native dependency missing.
	at org.jetbrains.skiko.Library.findAndLoad(Library.kt:105)
	at org.jetbrains.skiko.Library.load(Library.kt:59)
	at org.jetbrains.skia.impl.Library$Companion.staticLoad(Library.jvm.kt:12)
	at org.jetbrains.skia.Data.<clinit>(Data.kt:60)
	... 5 more
```

While the IDE sample works fine?

rock3r:
All fixed, thanks to @Walingar  🤟

Walingar (approved):

PR #133, opened 2023-09-26
Added logic that manages the scroll wheel click on the tabs

edivad1999:


rock3r:
Thanks for the first contribution @edivad1999 :) Daje

PR #134, opened 2023-09-26
Fabrizio.scarponi/111 spinning indicator circular progress indicator

fscarponi:
This commit adds a circular progress component to the UI that supports both small and large size icons. The base theme has been updated to provide styling to both standalone and ide The circular progress can be incorporated in the UI using the CircularProgress and CircularProgressBig components. This addition provides a more visually engaging way to show progress in the UI.


ide-laf-bridge/build.gradle.kts
```diff
@@ -7,6 +7,5 @@ plugins {
 dependencies {
     api(projects.intUi.intUiStandalone)
     compileOnly(libs.bundles.idea)
-
```
rock3r, line 10: Why?

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/IntUiBridge.kt
```diff
@@ -881,3 +885,43 @@ private fun readEditorTabStyle(iconData: IntelliJThemeIconData, svgLoader: SvgLo
         ),
     )
 }
+
+private fun readCircularProgressStyle(
+    svgLoader: SvgLoader,
+    iconData: IntelliJThemeIconData,
+): IntUiCircularProgressStyle =
```
rock3r, line 892: Nit: you don't need the explicit return type since it's a private function and you're not downcasting to, e.g., an interface

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/IntUiBridge.kt
```diff
@@ -881,3 +885,43 @@ private fun readEditorTabStyle(iconData: IntelliJThemeIconData, svgLoader: SvgLo
...
+    IntUiCircularProgressStyle(
+        frameTime = 125.milliseconds,
+        frameIcons = IntUiCircularProgressIcons(
+            listOf(
+                retrieveStatelessIcon("process/step_1.svg", svgLoader = svgLoader, iconData = iconData),
```
rock3r, line 897: This is the _old_ style one — and it is different from the one we use in the standalone. I don't think that we can use the stored icons in the bridge, we need to do the same thing we do in the standalone, since the IDE will build the SVG on the fly like we do. Well, we copy what the IDE does, but you get what I mean :D

int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiCircularProgressStyle.kt
```diff
@@ -0,0 +1,97 @@
...
+            ResourcePainterProvider.stateless(basePath = it, svgLoader = svgLoader)
+        }
+    }
+}
+
+@Composable
+fun intUiCircularProgressIcons(
+    svgLoader: SvgLoader,
+    iconPrefix: String = "icons/intui/animated/smallSpinner/",
+    iconsNames: List<String> = listOf(
```
rock3r, line 78: `iconNames`

core/src/main/kotlin/org/jetbrains/jewel/CircularProgressBar.kt
```diff
@@ -0,0 +1,41 @@
...
+    var currentFrame by remember { mutableStateOf(style.frameIcons.frames.first()) }
+
+    Icon(
+        modifier = modifier,
+        painter = currentFrame.getPainter(LocalResourceLoader.current).value,
+        contentDescription = null,
+    )
+
+    LaunchedEffect(Unit) {
+        while (true) {
```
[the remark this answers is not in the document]
fscarponi: When [LaunchedEffect](https://developer.android.com/jetpack/compose/side-effects) enters the Composition, it launches a coroutine with the block of code passed as a parameter. The coroutine will be cancelled if LaunchedEffect leaves the composition

ide-laf-bridge/build.gradle.kts
fscarponi: probably a lint rule or something like that when i started spam cmnd shift L, reverted!

rock3r (approved):

PR #153, opened 2023-10-03
Include only foundation-desktop in the published library

Walingar:
Common layout for Compose for Desktop libraries should be that "library" part depends only on needed foundation/material/ui parts and "sample" part depends on `compose.desktop.currentOs`, so dependency on skiko runtime will be provided for these samples.


rock3r (approved):
Thanks!

PR #165, opened 2023-10-10
Support different IntelliJ versions

Walingar:
This PR provides new modules ide-laf-bridge-232 and ide-laf-bridge-233 that will be used for platform specific parts that are changed during releases and cannot be used in common ide-laf-bridge.

ide-laf-bridge will depend on one of them depending on local property `supported.ij.version`

Publishing is also changed. Now all publications will have suffix `ij-$version` e.g. `0.7.1-ij-232`. When we support 233 we will have 2 publications e.g. `0.7.3-ij-233` and `0.7.3-ij-232` for all jewel parts.
In future it is better to split publishing process and publish everything except ide-laf-bridge with version without suffix and separately publish ide-laf-bridge and ide-laf-bridge-platform-specific with suffix depending on other parts without suffix

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,15 +5,9 @@ on:
   push:
     branches: [ main ]
 jobs:
-  publish:
-    name: Publish
+  publish-core:
+    name: Publish Jewel Core
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        supported-ij-version:
-          - 232
-#          - 233
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


Walingar:
Discussed with @rock3r that we will make fix it in this PR
> In future it is better to split publishing process and publish everything except ide-laf-bridge with version without suffix and separately publish ide-laf-bridge and ide-laf-bridge-platform-specific with suffix depending on other parts without suffix

buildSrc/src/main/kotlin/PublishConfiguration.kt
```diff
@@ -0,0 +1,35 @@
...
+                username = System.getenv("MAVEN_SPACE_USERNAME")
+                password = System.getenv("MAVEN_SPACE_PASSWORD")
+            }
+        }
+    }
+}
+
+internal fun MavenPom.configureJewelPom() {
+    name = "Jewel"
+    description = "intelliJ theming system in for Compose."
```
rock3r, line 22: Nit: can we fix this? I just noticed we've had a typo for a while :D

```suggestion
    description = "A theme for Compose for Desktop that implements the IntelliJ Platform look and feel."
```

.github/workflows/publish.yml
```diff
@@ -17,7 +18,33 @@ jobs:
...
+        with:
+          distribution: temurin
+          java-version: 17
+      - name: Setup Gradle
+        run: chmod +x gradlew
+      - name: Run Gradle
+        run: ./gradlew publishIdeMainPublicationsToSpaceRepository -Psupported.ij.version=${{ matrix.supported-ij-version }}
+        env:
+          MAVEN_SPACE_USERNAME: ${{secrets.MAVEN_SPACE_USERNAME}}
+          MAVEN_SPACE_PASSWORD: ${{secrets.MAVEN_SPACE_PASSWORD}}
```
rock3r, line 50: Nit: here and in other files, make sure there is an empty line at EOF. There is a convenient setting at _Settings | Editor | General_ that automates it:

image

rock3r (approved):

Issue #116, opened 2023-09-19
Replace reflection with public API access


rock3r:
We use reflection to access some data in the IJ 232 platform:
 * [ ] `patchIconPath` from `com.intellij.ui.icons.CachedImageIcon.kt`, used in `BridgeIconMapper`
 * [ ] `icons` and `iconColorsOnSelection` from `com.intellij.ide.ui.UITheme`, used in `org.jetbrains.jewel.bridge.UiThemeExtensions.kt`

The former has been made public in IJP 233, and we're waiting for an API to access the latter, which should hopefully be available in 233 as well. This issue is to track that we need to switch over.

Issue #131, opened 2023-09-22
Support IJ 23.3 EAP


rock3r:
Today the IJ 233 EAP was released, and this brings breaking changes to some of the internals we rely on.

Specifically:
 * `UITheme.colors` doesn't exist anymore
 * All `UITheme` private fields read from the JSON have been grouped in the `bean` field
 * The `LaFManager` hard-deprecated `getCurrentLookAndFeel()` in favour of `getCurrentUIThemeLookAndFeel()`

Possibly other things may be broken, too.

rock3r:
FYI @Walingar

rock3r:
For the record — I have started work on this, but will postpone it until we get some APIs we can interact with in the platform, so we can get rid of the hacks around reflection for 233+

rock3r:
The APIs we need are coming in EAP 3:

 * To get `UIThemeLookAndFeelInfo` — `LafManager.getCurrentUIThemeLookAndFeel()`
 * Then `UIThemeLookAndFeelInfo.describe()` will get you a `UIThemeExportableBean` that contains the info you need

PR #169, opened 2023-10-11
Support IJ 233 with new theming API

Walingar:
This PR splits implementations for 232 and 233, so that in 233 new theming API is used

Fixes: https://github.com/JetBrains/jewel/issues/116, https://github.com/JetBrains/jewel/issues/131


rock3r (approved):
Excellent!

PR #172, opened 2023-10-12
Handle non-existing JBColor key

Walingar:
Previous solution with FallbackColor didn't work since JBColor.namedColor always returns new instance and equals inside FallbackColor will return false in all cases


rock3r (approved):
It's arguably not a "better" way to do it than what we have now, and I agree it should be fixed in the platform, but I don't have anything against putting this in

Walingar:
@rock3r as I written in PR description current solution just do nothing, because equals always will return false.

Issue #83, opened 2023-07-08
Remove usage of data classes


rock3r:
They're problematic in terms of binary compat, we should use regular classes with equals and hashcode and toString implemented by hand. See [this article](https://jakewharton.com/public-api-challenges-in-kotlin/) for details.

rock3r:
Work is ongoing, started on 0.5.0, but will continue.

rock3r:
We can use something like https://github.com/drewhamilton/Poko to autogen equals/hashcode/tostring when not using data classes anymore

sushkov-pv:
Lack of the copy function makes overriding theme settings inconvenient. Are there plans to add copy functions for all theme classes?

rock3r:
Not currently, no. You're free to create your own, but we can't because they impact binary compatibility, as the article in the issue explains.

PR #173, opened 2023-10-12
Decorated window with custom titlebar support

devkanro:


core/src/main/kotlin/org/jetbrains/jewel/IconButton.kt
```diff
@@ -76,7 +76,9 @@ fun IconButton(
             .border(style.metrics.borderWidth, border),
         contentAlignment = Alignment.Center,
         content = {
-            content(buttonState)
+            onBackground(background) {
```
devkanro, line 79: Some trick for show dark icon on dark background in light theme (Light theme with dark title bar), you can find the impl below

core/src/main/kotlin/org/jetbrains/jewel/IntelliJTheme.kt
```diff
@@ -232,6 +236,10 @@ internal val LocalIsDarkTheme = staticCompositionLocalOf<Boolean> {
     error("No InDarkTheme provided")
 }
 
+internal val LocalOnDarkBackground = staticCompositionLocalOf<Boolean> {
```
devkanro, line 239: I created a new composition local and not to use the LocalIsDarkTheme because we didn't change the theme's properties.

core/src/main/kotlin/org/jetbrains/jewel/IntelliJTheme.kt
```diff
@@ -244,3 +252,14 @@ val LocalColorPalette = staticCompositionLocalOf<IntelliJThemeColorPalette> {
 val LocalIconData = staticCompositionLocalOf<IntelliJThemeIconData> {
     EmptyThemeIconData
 }
+
+@Composable
+fun onBackground(color: Color, content: @Composable () -> Unit) {
+    val locals = if (color.isSpecified && color.alpha == 1f) {
```
devkanro, line 258: Ignore transparent colors, because we have no context to blend colors and get it is dark or not.

core/src/main/kotlin/org/jetbrains/jewel/IntelliJTheme.kt
```diff
@@ -244,3 +252,14 @@ val LocalColorPalette = staticCompositionLocalOf<IntelliJThemeColorPalette> {
 val LocalIconData = staticCompositionLocalOf<IntelliJThemeIconData> {
     EmptyThemeIconData
 }
+
+@Composable
+fun onBackground(color: Color, content: @Composable () -> Unit) {
```
devkanro, line 257: Function may should be renamed, but I can't give a better name

core/src/main/kotlin/org/jetbrains/jewel/IntelliJThemeDefinition.kt
```diff
@@ -15,4 +16,8 @@ interface IntelliJThemeDefinition {
 
     val colorPalette: IntelliJThemeColorPalette
     val iconData: IntelliJThemeIconData
+
+    val extensionStyles: Array<ProvidedValue<*>>
```
devkanro, line 20: I created this for controls in external module like DecoratedWindow and TitleBar

core/src/main/kotlin/org/jetbrains/jewel/util/ColorExtensions.kt
```diff
@@ -20,3 +21,14 @@ fun Color.toHexString(): String {
         }
     }
 }
+
+fun Color.isDark(): Boolean {
```
devkanro, line 25: From IDEA

decorated-window/src/main/java/com/jetbrains/AccessibleAnnouncer.java
```diff
@@ -0,0 +1,45 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.jetbrains;
```
devkanro, line 24: JBR generated codes

core/src/main/kotlin/org/jetbrains/jewel/util/ColorExtensions.kt
```diff
@@ -20,3 +21,11 @@ fun Color.toHexString(): String {
         }
     }
 }
+
+fun Color.isDark(): Boolean = (luminance + 0.05) / 0.05 < 4.5
+
+val Color.luminance: Double
```
rock3r, line 27: Why not using the [`Color.luminance()`](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/Color#(androidx.compose.ui.graphics.Color).luminance()) method?

decorated-window/src/main/java/com/jetbrains/AccessibleAnnouncer.java
rock3r: Do we need everything? Can we slim it down to only what we actually need?

int-ui/int-ui-core/src/main/kotlin/org/jetbrains/jewel/intui/core/IntUiThemeDefinition.kt
```diff
@@ -9,16 +10,20 @@ import org.jetbrains.jewel.IntelliJThemeDefinition
 import org.jetbrains.jewel.IntelliJThemeIconData
 
 @Immutable
-class IntUiThemeDefinition(
+data class IntUiThemeDefinition(
```
rock3r, line 13: ❌  we can't use data classes, as a library, because they break binary compatibility. See #83

int-ui/int-ui-core/src/main/kotlin/org/jetbrains/jewel/intui/core/IntUiThemeDefinition.kt
```diff
@@ -9,16 +10,20 @@ import org.jetbrains.jewel.IntelliJThemeDefinition
...
     override val colorPalette: IntUiThemeColorPalette,
     override val iconData: IntelliJThemeIconData,
     override val globalMetrics: GlobalMetrics,
     override val defaultTextStyle: TextStyle,
     override val contentColor: Color,
+    override val extensionStyles: Array<ProvidedValue<*>> = emptyArray(),
 ) : IntelliJThemeDefinition {
 
+    override fun withExtensions(vararg extensions: ProvidedValue<*>): IntUiThemeDefinition =
+        copy(extensionStyles = extensionStyles + extensions)
```
rock3r, line 25: I was thinking, this could allow multiple identical `ProvidedValue`s to be passed in. Wouldn't it make sense to use a `Set` to back the extension styles, so only the latest addition is retained? Should be easier to debug.

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/IntUiTheme.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.jetbrains.jewel.intui.standalone.rememberSvgLoader
+import org.jetbrains.jewel.intui.window.styling.IntUiDecoratedWindowStyle
+import org.jetbrains.jewel.intui.window.styling.IntUiTitleBarStyle
+import org.jetbrains.jewel.window.styling.DecoratedWindowStyle
+import org.jetbrains.jewel.window.styling.LocalDecoratedWindowStyle
+import org.jetbrains.jewel.window.styling.LocalTitleBarStyle
+import org.jetbrains.jewel.window.styling.TitleBarStyle
+
+@Composable
+fun IntUiThemeDefinition.titleBarStyle(svgLoader: SvgLoader, lightHeaderInLight: Boolean = false): TitleBarStyle =
```
rock3r, line 16: I'm not a huge fan of taking the dark value from the theme and combining it with a boolean that is ignored if the theme is dark. What about making `TitleBarStyle` a sealed class that implements the various styles? It is a different pattern than we use elsewhere, but it's probably better for this specific case, as resulting API

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/styling/IntUiDecoratedWindowStyling.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Stable
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import org.jetbrains.jewel.window.styling.DecoratedWindowColors
+import org.jetbrains.jewel.window.styling.DecoratedWindowMetrics
+import org.jetbrains.jewel.window.styling.DecoratedWindowStyle
+
+@Stable data class IntUiDecoratedWindowStyle(
```
rock3r, line 12: Please do not use `data class`. See #83

rock3r: PS: code style requires annotations to be on a separate line.

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/styling/IntUiDecoratedWindowStyling.kt
```diff
@@ -0,0 +1,66 @@
...
+            metrics = intUiDecoratedWindowMetrics,
+        )
+
+        @Composable fun dark(
+            intUiDecoratedWindowColors: IntUiDecoratedWindowColors = IntUiDecoratedWindowColors.dark(),
+            intUiDecoratedWindowMetrics: IntUiDecoratedWindowMetrics = IntUiDecoratedWindowMetrics(1.dp),
+        ): IntUiDecoratedWindowStyle = IntUiDecoratedWindowStyle(
+            colors = intUiDecoratedWindowColors,
+            metrics = intUiDecoratedWindowMetrics,
+        )
```
rock3r, line 33: For consistency with the rest of Jewel:

```suggestion
        @Composable fun light(
            colors: IntUiDecoratedWindowColors = IntUiDecoratedWindowColors.light(),
            metrics: IntUiDecoratedWindowMetrics = IntUiDecoratedWindowMetrics(),
        ) = IntUiDecoratedWindowStyle(colors, metrics)

        @Composable fun dark(
            colors: IntUiDecoratedWindowColors = IntUiDecoratedWindowColors.dark(),
            metrics: IntUiDecoratedWindowMetrics = IntUiDecoratedWindowMetrics(),
        ) = IntUiDecoratedWindowStyle(colors, metrics)
```

(the metrics default value should be in the metrics constructor)

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/styling/IntUiDecoratedWindowStyling.kt
rock3r: PPS: this and all child classes can be `Immutable` since all the children contain only primitive types or `@Immutable` values (`Color` and `Dp`)

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/styling/IntUiTitleBarStyling.kt
```diff
@@ -0,0 +1,291 @@
...
+import org.jetbrains.jewel.styling.PainterProvider
+import org.jetbrains.jewel.styling.ResourcePainterProvider
+import org.jetbrains.jewel.styling.SimpleResourcePathPatcher
+import org.jetbrains.jewel.window.DecoratedWindowState
+import org.jetbrains.jewel.window.styling.TitleBarColors
+import org.jetbrains.jewel.window.styling.TitleBarIcons
+import org.jetbrains.jewel.window.styling.TitleBarMetrics
+import org.jetbrains.jewel.window.styling.TitleBarStyle
+
+@Stable data class IntUiTitleBarStyle(
```
rock3r, line 28: Same considerations as for `IntUiDecoratedWindowStyling`: code style, avoid data classes, etc

int-ui/int-ui-decorated-window/src/main/kotlin/org/jetbrains/jewel/intui/window/styling/IntUiTitleBarStyling.kt
```diff
@@ -0,0 +1,291 @@
...
+    override val colors: IntUiTitleBarColors,
+    override val metrics: IntUiTitleBarMetrics,
+    override val icons: TitleBarIcons,
+) : TitleBarStyle {
+
+    override fun iconButtonStyle(): IconButtonStyle {
+        return IntUiIconButtonStyle(
+            IntUiIconButtonColors(
```
rock3r, line 36: Please add parameter names, it's really hard to figure out what's what when all parameters have the same value

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -48,41 +55,81 @@ import org.jetbrains.jewel.samples.standalone.components.Tabs
...
-        title = "Jewel component catalog",
-        icon = icon,
-    ) {
+    application {
         var isDark by remember { mutableStateOf(false) }
+        var lightHeaderInLight by remember { mutableStateOf(false) }
         var swingCompat by remember { mutableStateOf(false) }
         val theme = if (isDark) IntUiTheme.darkThemeDefinition() else IntUiTheme.lightThemeDefinition()
+        val projectColor by remember {
+            derivedStateOf {
```
rock3r, line 72: We don't have backpressure problems, so there is no need to use `derivedStateOf`. See https://medium.com/androiddevelopers/jetpack-compose-when-should-i-use-derivedstateof-63ce7954c11b

rock3r (changes requested):

core/src/main/kotlin/org/jetbrains/jewel/IntelliJTheme.kt
```diff
@@ -244,3 +252,14 @@ val LocalColorPalette = staticCompositionLocalOf<IntelliJThemeColorPalette> {
...
+    val locals = if (color.isSpecified && color.alpha == 1f) {
+        arrayOf(LocalOnDarkBackground provides color.isDark())
+    } else {
+        emptyArray()
+    }
+
+    CompositionLocalProvider(values = locals, content)
```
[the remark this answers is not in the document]
devkanro: This is to avoid putting composable into conditional branches and avoid remembered state losed caused by conditional switching.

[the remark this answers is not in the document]
rock3r: Ah, I see. Yu have a good point.

decorated-window/src/main/java/com/jetbrains/AccessibleAnnouncer.java
devkanro: Removed all unused component

int-ui/int-ui-core/src/main/kotlin/org/jetbrains/jewel/intui/core/IntUiThemeDefinition.kt
devkanro: ProvidedValue is not comparable...I think

core/src/main/kotlin/org/jetbrains/jewel/FocusableComponentState.kt
```diff
@@ -16,11 +16,12 @@ interface FocusableComponentState : InteractiveComponentState {
         pressed: T,
         hovered: T,
         active: T,
+        isSwingCompatMode: Boolean = IntelliJTheme.isSwingCompatMode,
```
rock3r, line 19: Why is this passed in here? I don't want to have a way for components to ignore the compat mode implicitly — for those which it makes no sense, we have in other cases reimplemented the when. The idea is that this function should always do the right thing in the base case, which is to respect the flag.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -67,21 +71,28 @@ import java.net.URI
 fun main() {
     val icon = svgResource("icons/jewel-logo.svg")
     application {
-        var isDark by remember { mutableStateOf(false) }
-        var lightHeaderInLight by remember { mutableStateOf(false) }
+        var intTheme by remember { mutableStateOf(IntUiThemes.Light) }
```
rock3r, line 74: `intUiTheme`? `theme`?

core/src/main/kotlin/org/jetbrains/jewel/FocusableComponentState.kt
devkanro: Not all controls ignore hover and press events in swing compatibility mode. For example, DropDown in TitleBar has hover and press background.
I think the best way is to have each control handle the hover and press events by itself, like in devkanro/jewel#1

core/src/main/kotlin/org/jetbrains/jewel/styling/DropdownStyling.kt
```diff
@@ -41,6 +41,8 @@ interface DropdownColors {
             pressed = backgroundPressed,
             hovered = backgroundHovered,
             active = background,
+            // Dropdown has hovered/pressed in title bar
+            isSwingCompatMode = false,
```
devkanro, line 45: Check here @rock3r

core/src/main/kotlin/org/jetbrains/jewel/FocusableComponentState.kt
rock3r: I agree, what I was pointing to is that I'd rather keep this method unchanged, as components that want to behave differently can simply have their own, different implementation, instead of using this.

rock3r (approved):

PR #175, opened 2023-10-12
Improve error handling for missing 'supported.ij.version'

lamba92:
Updated 'supported.ij.version' property handling in IdeaConfiguration.kt to better manage cases when the property is not set. Instead of throwing an error, it now sets a default value and logs a warning message once, advising the user to provide the missing property.

### DIFF
--- a/buildSrc/src/main/kotlin/IdeaConfiguration.kt
+++ b/buildSrc/src/main/kotlin/IdeaConfiguration.kt
@@ -1,3 +1,4 @@
+import java.util.concurrent.atomic.AtomicBoolean
 import org.gradle.api.Project
 
 enum class SupportedIJVersion {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/IdeaConfiguration.kt
```diff
@@ -1,26 +1,37 @@
...
+        localProperty("supported.ij.version")
+            ?: rootProject.property("supported.ij.version")?.toString()
+    }.getOrNull()
+
+    if (prop == null) {
+        if (!warned.getAndSet(true)) {
+            logger.warn(
+                """
+                    No 'supported.ij.version' property provided. Falling back to IJ 233.
+                    It is recommended to provide it to avoid unexpected behavior.
```
rock3r, line 22:
```suggestion
                    It is recommended to provide it using local.properties file or -Psupported.ij.version to avoid unexpected behavior.
```

PR #176, opened 2023-10-12
Change plugin and artifactId in ide-laf-bridge modules

lamba92:
The `jewel-ij-publish` plugin has been removed from build.gradle.kts files in ide-laf-bridge-232 and ide-laf-bridge-233 modules, and replaced with the `jewel-publish` plugin. Also, the artifactId of publications was changed as per the respective module names.


Walingar (changes requested):
Current solution works this way (e.g. if we publish 0.7.4 version): 
1. it publishes all libraries except `ide-laf-bridge` and `ide-laf-bridge-platform-specific` with version 0.7.4
2. it publishes `ide-laf-bridge` with versions `0.7.4-ij-232` and `0.7.4-ij-233` using relevant ij builds for that (so binary compatibility will be ok) and this `ide-laf-bridge` will depend on `ide-laf-bridge-platform-specific:0.7.4-ij-232` and `ide-laf-bridge-platform-specific:0.7.4-ij-233` respectively.
3. It publishes `ide-laf-bridge-platform-specific` with versions `0.7.4-ij-232` and `0.7.4-ij-233` using relevant ij builds

So, in IJ we will depend on ide-laf-bridge:0.7.4-ij233 and everything will be alright.

Your solution will work another way:
1. It will publish all the libraries with version 0.7.4
2. It will publish `ide-laf-bridge-ij232` and `ide-laf-bridge-ij233` with version 0.7.4

And it won't be able to publish `ide-laf-bridge` 0.7.4 two times. Also, it won't work well, since `ide-laf-bridge:0.7.4` will depend either on ij232 either ij233
So, your PR is going to break proper dependencies

PR #177, opened 2023-10-12
Fix checkbox and radio button icons

devkanro:
This pr fix presentation of checkbox and radio button by moving icon to path that same as IDEA and the PaletteMapper can work fine with standalone


core/src/main/kotlin/org/jetbrains/jewel/IconMapper.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.compose.ui.res.ResourceLoader
+
+interface IconMapper {
+
+    fun mapPath(originalPath: String, iconData: IntelliJThemeIconData, resourceLoader: ResourceLoader): String
+}
+
+object IntelliJIconMapper : IconMapper {
+
+    private const val VERBOSE = true
```
Walingar, line 12: Let's disable VERBOSE by default

ide-laf-bridge/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
 }
 
 dependencies {
-    api(projects.intUi.intUiStandalone) {
+    api(projects.intUi.intUiCore) {
```
Walingar, line 10: Good point!

Walingar (approved):
Awesome, thank you!

core/src/main/kotlin/org/jetbrains/jewel/IconMapper.kt
devkanro: Done

PR #178, opened 2023-10-13
Publishing: SNAPSHOT suffix should always be last

Walingar:


rock3r (approved):
Lgtm but check with @lamba92 cause he's still working on related stuff

PR #181, opened 2023-10-17
Allow nullable IDs in TreeBuilder classes

fscarponi:
The `id` property in classes of the `TreeBuilder` has been made nullable. With this change, in such scenarios a `null` `id` can be added, with default fallback values assigned where necessary in the code. For example, in case of null `id` for a node or a leaf, it is either replaced by a combined string of its parent id and its index.


rock3r (approved):
:shipit:

PR #180, opened 2023-10-17
Refactor SvgLoader api

devkanro:
This PR aims to simplify the loading of icon resources, replacing the pain of having to pass SvgLoader and ResourceLoader everywhere.

The redesigned `PainterHint` abstracts the previous code that separates resource paths or svg palettes in different places.

The new `PainterProvider` and `ResourceResolver` provide a unified entry point for loading icon resources.

```Kotlin
val painterProvider =
    retrieveStatelessIcon("general/showInfos.svg", svgLoader, IntUiTheme.iconData)
val painter by painterProvider.getPainter(resourceLoader)
Icon(painter, contentDescription = null)
```
become more simple one like this:
```Kotlin
Icon("general/showInfos.svg", contentDescription = null)
```
or using PointerProvider like this:
```Kotlin
val painterProvider = rememberPainterProvider("general/showInfos.svg")
val painter by painterProvider.getPainter()
Icon(painter, contentDescription = null)
```
If your icons in other module, use custom ResourceResolver to load it:
```Kotlin
object MyResourceResolver: ClassLoaderResourceResolver(SomeClassInOtherModule::class.java.classLoader)

val painterProvider = rememberPainterProvider("general/showInfos.svg", MyResourceResolver)
val painter by painterProvider.getPainter()
Icon(painter, contentDescription = null)
```
Or use it in Icon component:
```Kotlin
Icon("general/showInfos.svg", contentDescription = null, resourceResolver = MyResourceResolver)
```

PR #188, opened 2023-10-17
Handle SVG loading exceptions and path patching issues

lamba92:
This commit adds error handling to SVG loading in the JewelSvgLoader, returning a default ColorPainter when the specified SVG resource cannot be loaded. It also addresses a breaking change in 233 EAP 4's path patching logic where it now returns a "reflective path". The change includes code that transforms this path back into the original path.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/jewel/JewelSvgLoader.kt
+++ b/core/src/main/kotlin/org/jetbrains/jewel/JewelSvgLoader.kt
@@ -3,6 +3,8 @@ package org.jetbrains.jewel
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.remember
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.painter.ColorPainter
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.ResourceLoader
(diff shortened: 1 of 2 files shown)


rock3r:
Re: the EAP 4 path patching workaround I would really, _really_ not have to put this horrible hack in the codebase. The fix needs to happen in the IJ platform, which shouldn't have changed the behaviour of `CachedImageIcon.kt#patchIconPath()` in such an incompatible way.

I'm ok with the rest of the patch, avoiding crashing when icons can't be loaded. We should consider a "strict mode" that crashes anyway, that could for example always be on when running from sources (so devs catch errors quickly)

devkanro:
These commits may conflict with #180 , after #180 is merged, I'll comment here and hint where the new logic is.

ide-laf-bridge/ide-laf-bridge-233/src/main/kotlin/org/jetbrains/jewel/bridge/IconsPathPatching.kt
```diff
@@ -21,6 +22,33 @@ fun getPatchedIconPath(
         val patchedPath = patchIconPath(originalPath.removePrefix("/"), classLoader)?.first
             ?: patchIconPath(fallbackPath, classLoader)?.first
 
+        // 233 EAP 4 broke path patching horribly, now it can return a
+        // "reflective path", which is a FQN to an ExpUIIcons entry.
+        // As a (hopefully) temporary solution, we undo this transformation
+        // back into the original path.
+        if (patchedPath?.startsWith("com.intellij.icons.ExpUiIcons") == true) {
```
devkanro, line 29: This file have been moved to `org.jetbrains.jewel.bridge.BridgeOverride` as an `PainterHint`.

core/src/main/kotlin/org/jetbrains/jewel/JewelSvgLoader.kt
```diff
@@ -37,8 +39,20 @@ class JewelSvgLoader(private val svgPatcher: SvgPatcher) : SvgLoader {
     ): Painter {
         val density = LocalDensity.current
 
-        val painter = useResource(resourcePath, loader) {
-            loadSvgPainter(it.patchColors(resourcePath), density)
+        val painter = try {
```
devkanro, line 42: Those logic have been moved to `org.jetbrains.jewel.painter.ResourcePainterProvider#loadPainter` in line 92

rock3r:
Superseded by #191

PR #190, opened 2023-10-18
Add tests for painter hint

### LINKED ISSUE #189: Add tests for PainterHint path patching behaviour
See https://github.com/JetBrains/jewel/pull/180#discussion_r1362510173

devkanro:
Close #189


core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -0,0 +1,122 @@
...
+        return result
+    }
+
+    @Test
+    fun `dark painter hint`() {
+        val basePath = "icons/github.svg"
+        basePath.applyPathHints(Dark(true)).let {
+            Assert.assertEquals("icons/github_dark.svg", it)
+        }
+    }
```
rock3r, line 58:
```suggestion
    @Test
    fun `dark painter hint should append suffix when isDark is true`() {
        val basePath = "icons/github.svg"
        val patchedPath = basePath.applyPathHints(Dark(true))
        Assert.assertEquals("icons/github_dark.svg", patchedPath)
    }

    @Test
    fun `dark painter hint should not append suffix when isDark is false`() {
        val basePath = "icons/github.svg"
        val patchedPath = basePath.applyPathHints(Dark(false))
        Assert.assertEquals(basePath, patchedPath)
    }
```

I would test all possible cases, here and for the other hints. For hints with a single boolean parameter that means checking both true and false cases. And in general, I'd try to have only one assertion per test case, because that makes it easier to figure out what gets broken :)

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -0,0 +1,122 @@
...
+        val basePath = "icons/checkbox.svg"
+        basePath.applyPathHints(Selected(true)).let {
+            Assert.assertEquals("icons/checkboxSelected.svg", it)
+        }
+    }
+
+    @Test
+    fun `size painter hint`() {
+        val basePath = "icons/github.svg"
+        basePath.applyPathHints(Size("20x20")).let {
```
rock3r, line 79: Should we also test cases where the Size gets an empty and/or blank string? I'd argue it should throw an error in that case. And for the `Int, Int` variant, we should also throw when the user passes zero or negative values, and have corresponding tests. What do you think?

devkanro: In my impl it will be ignored when pass empty string or null to Size

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+
+    private fun String.applyPaletteHints(vararg hints: PainterHint): String {
+        val doc = documentBuilderFactory.newDocumentBuilder().parse(this.toByteArray().inputStream())
+
+        hints.forEach {
+            when (it) {
+                is PainterSvgPatchHint -> it.patch(doc.documentElement)
+                else -> return@forEach
+            }
+        }
```
rock3r, line 69: Do you want to interrupt processing when it's not an SvgPatchHint? Or were you going for something like this?

```suggestion
        hints.filterIsInstance(PainterSvgPatchHint::class)
            .onEach { it.patch(doc.documentElement) }
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+    fun `size painter hint should not append suffix when size is empty or null`() {
+        val basePath = "icons/github.svg"
+        val patchedPath = basePath.applyPathHints(Size(""))
+        Assert.assertEquals(basePath, patchedPath)
+        val patchedPath2 = basePath.applyPathHints(Size(null))
+        Assert.assertEquals(basePath, patchedPath2)
+    }
+
+    @Test
+    fun `stateful painter hint should append Disable suffix when enable is false`() {
```
rock3r, line 133:
```suggestion
    fun `stateful painter hint should append Disabled suffix when enabled is false`() {
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
     }
 
     @Test
-    fun `stateful painter hint`() {
+    fun `size painter hint should not append suffix when size is empty or null`() {
+        val basePath = "icons/github.svg"
+        val patchedPath = basePath.applyPathHints(Size(""))
+        Assert.assertEquals(basePath, patchedPath)
+        val patchedPath2 = basePath.applyPathHints(Size(null))
+        Assert.assertEquals(basePath, patchedPath2)
```
rock3r, line 129: I would prefer if the inputs were validated and an error thrown in this case, but up to you

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
-            Assert.assertEquals("icons/checkboxPressed.svg", it)
-        }
-        basePath.applyPathHints(Stateful(state.copy(hovered = true))).let {
-            Assert.assertEquals("icons/checkboxHovered.svg", it)
-        }
     }
 
     @Test
-    fun `stroke painter hint`() {
+    fun `stateful painter hint disable state takes higher priority over other states`() {
```
rock3r, line 146:
```suggestion
    fun `stateful painter hint disabled state takes higher priority over other states`() {
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+                        hovered = true,
+                        focused = true,
+                    ),
+                ),
+            )
+        Assert.assertEquals("icons/checkboxDisabled.svg", patchedPath)
+    }
+
+    @Test
+    fun `stateful painter hint should append Focused suffix when focused is false`() {
```
rock3r, line 164:
```suggestion
    fun `stateful painter hint should append Focused suffix when focused is true`() {
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+    @Test
+    fun `stateful painter hint focused state takes higher priority over pressed and hovered states`() {
+        val basePath = "icons/checkbox.svg"
+        val state = CheckboxState.of(toggleableState = ToggleableState.Off)
+        val patchedPath = basePath.applyPathHints(Stateful(state.copy(pressed = true, hovered = true, focused = true)))
+        Assert.assertEquals("icons/checkboxFocused.svg", patchedPath)
+    }
+
+    @Test
+    fun `stateful painter hint should append Pressed suffix when pressed is false`() {
```
rock3r, line 180:
```suggestion
    fun `stateful painter hint should append Pressed suffix when pressed is true`() {
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+    @Test
+    fun `stateful painter hint pressed state takes higher priority over hovered state`() {
+        val basePath = "icons/checkbox.svg"
+        val state = CheckboxState.of(toggleableState = ToggleableState.Off)
+        val patchedPath = basePath.applyPathHints(Stateful(state.copy(pressed = true, hovered = true)))
+        Assert.assertEquals("icons/checkboxPressed.svg", patchedPath)
+    }
+
+    @Test
+    fun `stateful painter hint should append Hovered suffix when hovered is false`() {
```
rock3r, line 196:
```suggestion
    fun `stateful painter hint should append Hovered suffix when hovered is true`() {
```

core/src/test/kotlin/org/jetbrains/jewel/PainterHintTest.kt
```diff
@@ -49,74 +55,192 @@ class PainterHintTest : BasicJewelUiTest() {
...
+
+    @Test
+    fun `stroke painter hint should not append suffix when stroked is false`() {
+        val basePath = "icons/rerun.svg"
+        val patchedPath = basePath.applyPathHints(Stroke(false))
+        Assert.assertEquals(basePath, patchedPath)
+    }
+
+    @Test
+    fun `palette painter hint should patche colors in svg filled`() {
```
rock3r, line 218:
```suggestion
    fun `palette painter hint should patch colors correctly in SVG`() {
```

rock3r (approved):

rock3r:
Great stuff, I've proposed some test name improvements

PR #191, opened 2023-10-19
Handle SVG loading exceptions and path patching issues in IJ 233 EAP 4

rock3r:
Same as #188 but up-to-date compared to main.


core/src/main/kotlin/org/jetbrains/jewel/painter/ResourcePainterProvider.kt
```diff
@@ -155,6 +159,23 @@ class ResourcePainterProvider(
         }
     }
 
+    @Composable
+    private fun createVectorDrawablePainter(url: URL, density: Density): VectorPainter {
+        val vector = url.openStream().use {
```
devkanro, line 164: Should we fallback to Magenta for bitmap and xml too?

core/src/main/kotlin/org/jetbrains/jewel/painter/ResourcePainterProvider.kt
```diff
@@ -70,56 +76,31 @@ class ResourcePainterProvider(
...
                 }
                 add(path)
             }
         }.reversed()
 
-        val url = pathStack.firstNotNullOfOrNull {
-            resolveResource(it)
-        } ?: error("Resource '$basePath(${hints.joinToString()})' not found")
+        val url = pathStack.firstNotNullOfOrNull { resolveResource(it) }
+            ?: error("Resource '$basePath(${hints.joinToString()})' not found")
```
devkanro, line 95: Should we fallback to Magenta for unresolvable resource too?

devkanro (commented):
Generally good, but the rules about fallback may still need to be discussed

core/src/main/kotlin/org/jetbrains/jewel/painter/ResourcePainterProvider.kt
rock3r: You've got a good point. Will do!

core/src/main/kotlin/org/jetbrains/jewel/painter/ResourcePainterProvider.kt
rock3r: You've got a good point. Will do!

core/src/main/kotlin/org/jetbrains/jewel/painter/ResourcePainterProvider.kt
```diff
@@ -70,56 +81,32 @@ class ResourcePainterProvider(
...
```
devkanro, line 100: It still throw error in Non-Debug mode

devkanro (approved):
LGTM

lamba92 (approved):

PR #192, opened 2023-10-19
The big refactor — step 1: prep work

### LINKED ISSUE #83: Remove usage of data classes

rock3r:
This is the first step of "the big refactor".

In this step:
 * We replace data classes in public APIs with [Poko](https://github.com/drewhamilton/Poko) — a compiler plugin that generates `equals`/`hashcode`/`toString` but not methods like `copy` and `componentN` which easily break binary compatibility. This fixes #83 (see issue for more details)
 * We add the kotlinx binary-compatibility-validator plugin, that generates a list of all public APIs exposed by the published modules — the bulk of the changes in this PR are the new `.api` files, that track the current state of the public API (they can be skipped when reviewing). A new `apiCheck` is added to the `check` task that ensures no changes break binary compatibility by mistake.
 * We add a new `validatePublicApi` task to published modules that breaks the `check` task whenever a data class is detected in the public API

---

> [!IMPORTANT]
> After this PR is merged, whenever adding new APIs, you need to run the `apiDump` method to update the `.api` files. If API changes are detected that are not in the `.api` files, the CI will fail.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/jewel/foundation/Stroke.kt
+++ b/core/src/main/kotlin/org/jetbrains/jewel/foundation/Stroke.kt
@@ -6,6 +6,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.graphics.isUnspecified
 import androidx.compose.ui.unit.Dp
+import dev.drewhamilton.poko.Poko
 
 sealed class Stroke {
     @Immutable
@@ -15,7 +16,8 @@ sealed class Stroke {
     }
 
     @Immutable
-    data class Solid internal constructor(
+    @Poko
+    class Solid internal constructor(
         val width: Dp,
         val color: Color,
         val alignment: Alignment,
@@ -23,7 +25,8 @@ sealed class Stroke {
     ) : Stroke()
 
     @Immutable
-    data class Brush internal constructor(
+    @Poko
+    class Brush internal constructor(
         val width: Dp,
         val brush: androidx.compose.ui.graphics.Brush,
         val alignment: Alignment,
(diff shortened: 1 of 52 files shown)


core/src/main/kotlin/org/jetbrains/jewel/foundation/Stroke.kt
```diff
@@ -6,6 +6,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.graphics.isUnspecified
 import androidx.compose.ui.unit.Dp
+import dev.drewhamilton.poko.Poko
```
devkanro, line 9: Maybe we can custom this annotation, like `org.jetbrains.jewel.Data` or other something

devkanro:
> After this PR is merged, whenever adding new APIs, you need to run the apiDump method to update the .api files

And how about add a CI pipeline to make sure contributors updated the api files.

rock3r:
@devkanro if you forget to update the API signatures, the ci will fail

core/src/main/kotlin/org/jetbrains/jewel/foundation/Stroke.kt
rock3r: We can, but I don't think it's really necessary

rock3r: Done!

devkanro (approved):
LGTM

PR #193, opened 2023-10-20
The big refactor — part 2

rock3r:
This is the second part of the big refactor. It removes all the unnecessary interfaces and makes creating the standalone versions much easier.

This is a huge refactoring, but I have tested it as much as possible. We may have some issues slip through, but we're planning a bugfix 0.8.1 release by next week in case.


devkanro (approved):
Nicely done! but the CI broken...

PR #195, opened 2023-10-20
The big refactor — part 3: more interfaces

rock3r:
This PR flattens more interfaces and cleans up more code. Unless I missed something, this is the almost the end of the first Big Refactor™! This required updating the theme generator as well, since the FQNs and structures involved changed. Expect a lot of "more of the same" as #193 :)

After this PR, there are only two small things left (with large impact but no real code changes):

1. Reorganising packages across modules
2. Splitting out the jewel-foundation module from jewel-core


devkanro (approved):
LGTM, nicely done

PR #197, opened 2023-10-21
Badge painter hint support

devkanro:
截屏2023-10-21 21 38 13

### DIFF
--- a/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/Icons.kt
+++ b/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/Icons.kt
@@ -15,6 +15,8 @@ import androidx.compose.ui.unit.dp
 import org.jetbrains.jewel.samples.standalone.StandaloneSampleIcons
 import org.jetbrains.jewel.ui.component.GroupHeader
 import org.jetbrains.jewel.ui.component.Icon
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/org/jetbrains/jewel/painter/badge/DotBadgeShape.kt
```diff
@@ -0,0 +1,64 @@
...
+) : BadgeShape {
+
+    override fun createHoleOutline(size: Size, layoutDirection: LayoutDirection, density: Density): Outline =
+        createOutline(size, layoutDirection, density, hole = true)
+
+    override fun createOutline(size: Size, layoutDirection: LayoutDirection, density: Density): Outline =
+        createOutline(size, layoutDirection, density, hole = false)
+
+    private fun createOutline(size: Size, layoutDirection: LayoutDirection, density: Density, hole: Boolean): Outline {
+        val dotSize = size.width.coerceAtMost(size.height)
```
rock3r, line 29: If this is supposed to be circular, why not taking a size as `Dp` instead of `Size`?

rock3r: Explained in private

rock3r (approved):

rock3r (approved):
Great job! Thanks

PR #201, opened 2023-10-23
Make lastActiveItemIndex public, so selection on hover can be implemented

Walingar:
For the case of selection on hover it is needed to manually modify lastActiveItemIndex, so let's make it public


fscarponi (approved):

fscarponi (commented):
probably expose a getter will be safer… but ok if is a temporary solution wating for the refactoring for items state

Walingar:
@fscarponi this is a Kotlin property, so it's always getter and setter functions

fscarponi:
> @fscarponi this is a Kotlin property, so it's always getter and setter functions

did you need to set this property from the outside? 
if not, you should expose the property getter but not the setter

PR #202, opened 2023-10-24
Automatically check for IJ Platform updates

rock3r:
This PR adds a new plugin into buildSrc that queries the JetBrains APIs to determine the latest IJ Platform available, and errors out when the version we're using is not the latest available. It keeps into consideration that if we're on a stable platform version, we don't want to upgrade to a non-stable one, too.

This PR additionally sets up a CI job that checks daily (at 8 AM in _some_ timezone lol) that there are no available updates. Right now, this is going to fail both for 232 and 233, but I will update the platforms in a separate PR, right after this.

Example failure for 232:

image


.github/workflows/check-ide-version.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: zulu
+          java-version: 17
+          cache: gradle
+      - name: Setup Gradle
+        run: chmod +x gradlew
+      - name: Run Gradle
+        run: ./gradlew checkLatestIntelliJPlatformBuild -Psupported.ij.version=${{ matrix.supported-ij-version }}
```
devkanro, line 29: Maybe we can create a PR automatically for updaing the IJ version, like the dep bot of GitHub

.github/workflows/publish.yml
```diff
@@ -15,8 +15,9 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
-          distribution: temurin
+          distribution: zulu
```
devkanro, line 18: Just curious, why was the JDK distribution modified?

.github/workflows/check-ide-version.yml
rock3r: Could be a next step, but until we have solid testing in place I would avoid risking e.g., the bridge being broken because some API got changed in the platform. For now we should always do manual smoke testing when updating

.github/workflows/publish.yml
rock3r: Aligned it with the rest of the ci jobs. Felt weird running checks on one and building the artefacts with the other

lamba92 (approved):

PR #204, opened 2023-10-24
Cleanup IDE demo

rock3r:
* Use IconButton (remove todo)
* Make IconButton behave like Actions that pop up a menu (they stay pressed when the menu is open)
* Make RightColumn and LeftColumn private as they should

### DIFF
--- a/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
+++ b/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
@@ -10,6 +10,7 @@ import androidx.compose.animation.core.tween
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.ScrollState
 import androidx.compose.foundation.background
+import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.interaction.HoverInteraction
 import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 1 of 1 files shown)


samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
```diff
@@ -352,26 +353,29 @@ private fun OverflowMenu(
...
-        contentDescription = "Select data source",
-        modifier = Modifier
-            .fillMaxHeight()
-            .clickable(interactionSource, null) { menuVisible = !menuVisible }
-            .background(backgroundColor, RoundedCornerShape((DarculaUIUtil.BUTTON_ARC.unscaled / 2).dp)),
-    )
+    // Emulates Swing actions that pop up menus — they stay pressed while the menu is open
+    IconButton(
+        modifier = Modifier.fillMaxHeight()
+            .thenIf(menuVisible) {
```
devkanro, line 361: Should we provide forcepressed attribute to IconButton?

devkanro (approved):

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
rock3r: I was considering we should have a state-exposing version of all components instead, as a more general purpose approach, but would like to see more use cases first

PR #213, opened 2023-10-28
Stroke hint, ExtensionComponentStyling api and Unfocusable IconButton/Tab

devkanro:
This PR contains 3 changes:

- Support Stroke hint via SVG pacthing, this commit also includes some refactoring of Painter Hints, which now provides context information for Hint implementations through PainterProviderScope
- Refactor Component Styling API to fix extension style is not updated in #210 
- Make IconButton and Tag not focuable

Screenshots for Stroke hint:
截屏2023-10-29 00 26 33


rock3r (approved):

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/MenuStyling.kt
```diff
@@ -32,6 +33,7 @@ class MenuColors(
     val border: Color,
     val shadow: Color,
     val itemColors: MenuItemColors,
+    val isDark: Boolean = background.isDark(),
```
rock3r, line 36: This here is weird, we should only ever depend on the `JewelTheme.isDark` value for this kind of check

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
```diff
@@ -34,14 +40,70 @@ fun IconButton(
...
+        content = content,
+    )
+}
+
+@Composable
+private fun IconButtonImpl(
+    selected: Boolean,
+    modifier: Modifier,
+    enabled: Boolean,
+    canActivate: Boolean,
```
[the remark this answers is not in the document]
devkanro: We need change IconButtonState in `onActivated`, but we can't change it outside IconButtonImpl

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/MenuStyling.kt
devkanro: I know it's a little weird, but this is used for Dropdown menu in Light theme with Dark header.

- Main(Light)
  - TitleBar(Dark)
    - Dropdown(Dark)
      - DropdownMenu(Light)

Maybe we should provide isDark for all components based on popup, its isDark could be diff with main area.

rock3r: But the whole point of the `OverrideDarkTheme` composable is to set the appropriate dark theme. If you're overriding the dark-ness of the menu or whatever, you should provide an appropriate styling at that site (by composition locals), instead of having flags in the styling

rock3r: For example, in pseudocode:

```kotlin
OverrideDarkTheme(isDark = true) {
    CompositionLocalProvider(
        LocalMenuStyling provides ...,
    ) {
        // ...
    }        
}
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/PainterProviderScope.kt
```diff
@@ -1,6 +1,8 @@
 package org.jetbrains.jewel.ui.painter
 
-interface PainterProviderScope {
+import androidx.compose.ui.unit.Density
+
+interface PainterProviderScope : Density {
```
rock3r, line 5: Why extending it instead of having it as a property? Curious about the reasoning, not a blocker.

rock3r (approved):

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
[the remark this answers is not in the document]
rock3r: What if the state was passed in to the impl as well?

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/PainterProviderScope.kt
devkanro: This is convenient for using functions like toDp in Hint where you don't need with(density)

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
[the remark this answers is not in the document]
devkanro: In that case, we need to maintain state outside of Impl, which may cause these codes to be written twice in IconButton and SelectableIconButton, but in fact these codes are also the main content of Impl.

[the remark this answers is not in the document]
devkanro: Remember state outside and pass state into Impl

```kotlin
fun IconButton() {
    val state = remember { mutableStateOf(IconButtonState.of()) }

    IconButtonImpl(state)
}
```

PR #215, opened 2023-10-30
feat: Add expandTree() and collapseTree() methods to TreeState

fscarponi:
`fun expandTree()` `fun collapseTree()` are now available in the `TreeState` to respectively expand and collapse the tree nodes


rock3r (approved):
LGTM, but you need to run `apiDump` when you change in any way the public API

PR #216, opened 2023-10-30
Update SelectableLazyColumn for key separation and update Gradle wrapper

lamba92:
Refactor SelectableLazyColumn and related classes, separating `selectionKey` and `uiKey`. `selectionKey` is used for selection and `uiKey` triggers recompositions. Also, upgraded Gradle wrapper to 8.4. This change improves re-composition handling and provides an easier way to control item selection without unnecessary recompositions.

### DIFF
--- a/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
+++ b/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
@@ -1,6 +1,5 @@
 package org.jetbrains.jewel.foundation.lazy
 
-import androidx.compose.foundation.focusGroup
 import androidx.compose.foundation.focusable
 import androidx.compose.foundation.gestures.FlingBehavior
 import androidx.compose.foundation.gestures.ScrollableDefaults
@@ -73,10 +72,7 @@ fun SelectableLazyColumn(
     val focusRequester = remember { FocusRequester() }
     LazyColumn(
         modifier = modifier
-            .onFocusChanged {
-                isFocused = it.isFocused || it.hasFocus
-            }
-            .focusGroup()
+            .onFocusChanged { isFocused = it.hasFocus }
             .focusRequester(focusRequester)
             .focusable(interactionSource = interactionSource)
             .onPreviewKeyEvent { event ->
--- a/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
+++ b/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.jewel.foundation.lazy
 
+import androidx.compose.foundation.focusGroup
 import androidx.compose.foundation.focusable
 import androidx.compose.foundation.gestures.FlingBehavior
 import androidx.compose.foundation.gestures.ScrollableDefaults
@@ -54,13 +55,13 @@ fun SelectableLazyColumn(
     val container = SelectableLazyListScopeContainer()
         .apply(content)
 
-    val keys = remember(container) {
-        container.getKeys()
+    val selectionKeys = remember(container) {
+        container.getSelectionKeys()
     }
     var isFocused by remember { mutableStateOf(false) }
 
     fun evaluateIndexes(): List<Int> {
-        val keyToIndexMap = keys.withIndex().associateBy({ it.value.key }, { it.index })
+        val keyToIndexMap = selectionKeys.withIndex().associateBy({ it.value.value }, { it.index })
         return state.selectedKeys
             .mapNotNull { selected -> keyToIndexMap[selected] }
             .sorted()
@@ -72,13 +73,16 @@ fun SelectableLazyColumn(
     val focusRequester = remember { FocusRequester() }
     LazyColumn(
         modifier = modifier
-            .onFocusChanged { isFocused = it.hasFocus }
+            .onFocusChanged {
+                isFocused = it.isFocused || it.hasFocus
+            }
+            .focusGroup()
             .focusRequester(focusRequester)
             .focusable(interactionSource = interactionSource)
             .onPreviewKeyEvent { event ->
                 if (state.lastActiveItemIndex != null) {
                     val actionHandled = keyActions
-                        .handleOnKeyEvent(event, keys, state, selectionMode)
+                        .handleOnKeyEvent(event, selectionKeys, state, selectionMode)
                         .invoke(event)
                     if (actionHandled) {
                         scope.launch {
@@ -99,12 +103,12 @@ fun SelectableLazyColumn(
     ) {
         container.getEntries().forEach { entry ->
             when (entry) {
-                is Entry.Item -> item(entry.key, entry.contentType) {
+                is Entry.Item -> item(entry.uiKey, entry.contentType) {
                     val itemScope = SelectableLazyItemScope(
-                        isSelected = entry.key in state.selectedKeys,
+                        isSelected = entry.selectionKey in state.selectedKeys,
                         isActive = isFocused,
                     )
-                    if (keys.any { it.key == entry.key && it is SelectableLazyListKey.Selectable }) {
+                    if (selectionKeys.any { it.value == entry.selectionKey && it is SelectableLazyListKey.Selectable }) {
                         Box(
                             modifier = Modifier.selectable(
                                 requester = focusRequester,
@@ -112,8 +116,8 @@ fun SelectableLazyColumn(
                                 actionHandler = pointerEventActions,
                                 selectionMode = selectionMode,
                                 selectableState = state,
-                                allKeys = keys,
-                                itemKey = entry.key,
+                                allKeys = selectionKeys,
+                                itemKey = entry.selectionKey,
                             ),
                         ) {
                             entry.content.invoke(itemScope)
@@ -125,11 +129,12 @@ fun SelectableLazyColumn(
 
                 is Entry.Items -> items(
                     count = entry.count,
-                    key = { entry.key(entry.startIndex + it) },
+                    key = { entry.uiKey(entry.startIndex + it) },
                     contentType = { entry.contentType(it) },
                 ) { index ->
-                    val itemScope = SelectableLazyItemScope(entry.key(index) in state.selectedKeys, isFocused)
-                    if (keys.any { it.key == entry.key(index) && it is SelectableLazyListKey.Selectable }) {
+                    val itemScope =
+                        SelectableLazyItemScope(entry.selectionKey(index) in state.selectedKeys, isFocused)
+                    if (selectionKeys.any { it.value == entry.selectionKey(index) && it is SelectableLazyListKey.Selectable }) {
                         Box(
                             modifier = Modifier.selectable(
                                 requester = focusRequester,
@@ -137,8 +142,8 @@ fun SelectableLazyColumn(
                                 actionHandler = pointerEventActions,
                                 selectionMode = selectionMode,
                                 selectableState = state,
-                                allKeys = keys,
-                                itemKey = entry.key(index),
+                                allKeys = selectionKeys,
+                                itemKey = entry.selectionKey(index),
                             ),
                         ) {
                             entry.itemContent.invoke(itemScope, index)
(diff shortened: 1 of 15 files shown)


settings.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencyResolutionManagement {
 
 plugins {
     id("com.gradle.enterprise") version "3.15.1"
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.7.0"
```
rock3r, line 30: Please avoid unrelated changes.

rock3r (changes requested):

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -72,13 +73,16 @@ fun SelectableLazyColumn(
     val focusRequester = remember { FocusRequester() }
     LazyColumn(
         modifier = modifier
-            .onFocusChanged { isFocused = it.hasFocus }
+            .onFocusChanged {
+                isFocused = it.isFocused || it.hasFocus
+            }
+            .focusGroup()
```
Walingar, line 79: There shouldn't be any focusable elements inside SelectableLazyColumn, otherwise `tab` behavior will work incorrectly, so we need to check only isFocused state and don't make it focusGroup

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -40,23 +48,28 @@ interface SelectableLazyListScope {
...
+        selectionKey: (index: Int) -> Any,
+        uiKey: (index: Int) -> Any = selectionKey,
         contentType: (index: Int) -> Any? = { null },
         selectable: (index: Int) -> Boolean = { true },
         itemContent: @Composable SelectableLazyItemScope.(index: Int) -> Unit,
     )
 
     /**
      * A method that enables sticky header behavior in a list or grid view.
+     * By default `[selectionKey] = [uiKey]`, but you can assign [uiKey] if you need to
```
Walingar, line 60: But if we keep only uiKey as it was before, then with unchanged uiKey element can be recomposed if State is changed. 
Key is used mostly for keeping scroll in a proper position on items modification, not for skiping recomposition in this case

Can you please provide a usecase where you need different uiKey and selectionKey? Since example for Chips and Tree is wrong, since you need to provide not string but objects that hold State<String> if you want to change it.

Walingar (changes requested):
I don't see a reason for this change, so for now I am against of providing complicated logic

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
fscarponi: Probably I'm missing something here, the only selectable element is the LazyColumn (due SEL is a wrapper).
if you are going to remove focusable from the LC, then you will have no way to send keybindings event to the component. 
BTW in most use cases, I doubt that users will not put something `clickable` inside the SLC, and if I'm not mind wrong, we cannot prevent that
Probably we can accept that and find a way to handle `TAB` key-event properly by moving in some way the focus to the next element

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
fscarponi: > Can you please provide a usecase where you need different uiKey and selectionKey
I can try to give you a reason: imagine you have a not so small item in the dataset, 
- probably you will handle the item key with a minimal key (as example, you can use a UUID as key)
- now something inside this item will change (or the whole item to)
- if you have the same key for both, your item will not be shown in the list; otherwise, you lose the selection because the key is not the same

Walingar: Sorry, I still don't get it. Let's just use selectionId as itemId then? It's the same.

> you lose the selection because the key is not the same

And this is expected behaviour, item is not the same, it shouldn't be selected. For sure, it depends on requirements, but this should be handled by SelectableLazyColumn itself then. I mean how to handle selection when selected items are removed

lamba92: It is needed in case you want to decouple the selection from recompositions. Indeed having a state inside the item to pass would be optimal, but in the case of PKGS it is not possible. We have a huge complex model and all the data to construct the tree is not reactive (flows/states ecc...). Everytime there is a change in a single part of the model, the entirety of the model gets re-emitted.

We could transform a portion of the model from plain properties to reactive components but that would require a huge rewrite and would also mean a lot of issues with the IJ platform (not everything is easily transformed into a reactive component).

Walingar: So then you need to introduce ids for your items and use them as keys

Walingar: Otherwise, scrolling will be broken on recomposition also.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -11,26 +11,34 @@ interface SelectableLazyListScope {
...
+     * @param uiKey The unique identifier for the item used for triggering recompositions.
      * @param contentType The type of content displayed in the item.
      * @param selectable Determines if the item is selectable. Default is
      *     `true`.
      * @param content The content of the item as a composable function.
      */
     fun item(
-        key: Any,
+        selectionKey: Any,
+        uiKey: Any = selectionKey,
```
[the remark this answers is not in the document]
lamba92: What about `selectionKey` and `itemKey`?  I mean the first is for the selection, the second is passed to `LazyColumn` and there is called `itemKey`.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -174,8 +197,8 @@ fun <T : Any> SelectableLazyListScope.itemsIndexed(
 
 @Composable
 fun LazyItemScope.SelectableLazyItemScope(
-    isSelected: Boolean = false,
-    isActive: Boolean = false,
+    isSelected: Boolean,
+    isActive: Boolean,
```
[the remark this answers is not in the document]
lamba92: Nope, this PR is not about that

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BuildTree.kt
```diff
@@ -8,33 +8,45 @@ fun <T> buildTree(builder: TreeBuilder<T>.() -> Unit): Tree<T> =
     TreeBuilder<T>().apply(builder).build()
 
 class TreeBuilder<T> : TreeGeneratorScope<T> {
-    sealed class Element<T> {
+    sealed interface Element<T> {
 
-        abstract val id: Any?
+        val data: T
+        val selectionId: Any?
+        val uiId: Any?
```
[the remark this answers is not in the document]
lamba92: up

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/ChipsAndTree.kt
```diff
@@ -55,36 +56,41 @@ fun ChipsAndTree() {
...
+    var listOfItems by remember {
+        mutableStateOf(List((5000..10000).random()) { "Item $it" })
+    }
+    LaunchedEffect(Unit) {
+        while (true) {
+            delay(1.seconds)
+            listOfItems = listOfItems.mapIndexed { index, s ->
+                if (index != 2) s else "WHOOO ${Random.nextInt(0, 10)}"
+            }
+        }
```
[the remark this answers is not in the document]
lamba92: It is a test for the changes. The objective of this PR is to allow to recompose when an item changes without loosing the selection. In other words, the selection should be on a subset of the data available in the item.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/ChipsAndTree.kt
```diff
@@ -55,36 +56,41 @@ fun ChipsAndTree() {
...
-                    Modifier.then(
+        modifier = Modifier.size(200.dp, 200.dp),
+        onSelectedIndexesChanged = { indexes ->
+            println(indexes.joinToString { listOfItems[it] })
+        },
+    ) {
+        items(
+            count = listOfItems.size,
+            selectionKey = { index -> index },
+            uiKey = { index -> listOfItems[index] },
```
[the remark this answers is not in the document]
lamba92: The objective of this PR is to allow to recompose when an item changes without loosing the selection. In other words, the selection should be on a subset of the data available in the item.

settings.gradle.kts
lamba92: It is necessary to run reliably the sample. I will move it to another PR

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -11,26 +11,34 @@ interface SelectableLazyListScope {
 
     /**
      * Represents an item in a selectable lazy list.
+     * By default `[selectionKey] = [uiKey]`, but you can assign [uiKey] if you need to
```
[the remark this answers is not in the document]
lamba92: removed!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-all.zip
```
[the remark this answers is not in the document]
lamba92: reverted, see [PR#231](https://github.com/JetBrains/jewel/pull/231)

samples/standalone/build.gradle.kts
```diff
@@ -37,3 +37,14 @@ compose.desktop {
...
+afterEvaluate {
+    tasks {
+        named<JavaExec>("run") {
+            javaLauncher = project.javaToolchains.launcherFor {
+                languageVersion = JavaLanguageVersion.of(17)
+                vendor = JvmVendorSpec.JETBRAINS
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
lamba92: reverted, see [PR#231](https://github.com/JetBrains/jewel/pull/231)

settings.gradle.kts
lamba92: reverted, see [PR#231](https://github.com/JetBrains/jewel/pull/231)

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
lamba92: I think I agree with @Walingar. On the other hand we need to warn the user to not use focusable or clickable modifiers on items inside the list. A solution could be to add such modifiers with all their overloads as extension members of `SelectableLazyListScope` and mark such modifiers as errors using annotations.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/ChipsAndTree.kt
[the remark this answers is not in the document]
rock3r: Yeah it's not something I would include in the sample for everyone though, especially since it's not something we want anyone to _actually_ be doing :P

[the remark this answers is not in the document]
rock3r: Please only keep this change locally while you develop.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/components/ChipsAndTree.kt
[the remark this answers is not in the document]
rock3r: > the selection should be on a subset of the data available in the item

And yet here the selection is _not_ based on the item data 🤔

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
rock3r: I don't know how that scales. I reckon a Lint check would the right way to go, but since we don't have that, a Detekt check would be good as well.

PR #231, opened 2023-10-31
Update Gradle version and Java toolchain

lamba92:
Updated Gradle version in gradle-wrapper.properties from 8.3 to 8.4. Also, reconfigured the Java and Kotlin JVM toolchains to use JetBrains JDK 17. In addition, added a new plugin "foojay-resolver-convention" in settings.gradle.kts.

### DIFF
--- a/samples/standalone/build.gradle.kts
+++ b/samples/standalone/build.gradle.kts
@@ -37,3 +37,17 @@ compose.desktop {
         }
     }
 }
+
+tasks {
+    withType<JavaExec> {
+        // afterEvaluate is needed because the Compose Gradle Plugin
+        // register the task in the afterEvaluate block
+        afterEvaluate {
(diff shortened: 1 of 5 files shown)


samples/standalone/build.gradle.kts
```diff
@@ -37,3 +37,17 @@ compose.desktop {
         }
     }
 }
+
+tasks {
+    withType<JavaExec> {
+        // afterEvaluate is needed because the Compose Gradle Plugin
+        // register the task in the afterEvaluate block
```
rock3r, line 44: Worth asking for a fix in the Compose Gradle plugin?

rock3r (approved):

samples/standalone/build.gradle.kts
lamba92: Yes, but in the meantime this is needed for us.

PR #234, opened 2023-11-01
Support IDE zoom and presentation mode

devkanro:
截屏2023-11-01 10 34 02

PR #239, opened 2023-11-02
Fix icon don't be scaled after density changed

### LINKED ISSUE #236: Icons do not scale dynamically in IDE
See [here](https://github.com/JetBrains/jewel/issues/217#issuecomment-1787645809). 

Icons do scale when you recompose them tho.

devkanro:
截屏2023-11-02 11 19 14

Close #236 and resolve comments in #234

### DIFF
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
@@ -35,7 +35,6 @@ import org.jetbrains.skiko.awt.font.AwtFontManager
 import org.jetbrains.skiko.toSkikoTypefaceOrNull
 import java.awt.Dimension
 import java.awt.Font
-import java.awt.GraphicsEnvironment
 import java.awt.Insets
 import javax.swing.UIManager
 
@@ -222,13 +221,9 @@ internal operator fun TextUnit.plus(delta: Float): TextUnit =
         else -> this
     }
 
-internal fun retrieveDensity(): Density {
+internal fun retrieveIdeaDensity(sourceDensity: Density): Density {
     val ideaScale = UISettingsUtils.getInstance().currentIdeScale
-    val scale = GraphicsEnvironment.getLocalGraphicsEnvironment()
-        .defaultScreenDevice
-        .defaultConfiguration
-        .defaultTransform
-        .scaleX * ideaScale
+    val scale = ideaScale * sourceDensity.density
 
-    return Density(scale.toFloat(), 1f)
+    return Density(scale, sourceDensity.fontScale)
 }
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
@@ -1,0 +1,41 @@
+package org.jetbrains.jewel.bridge
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.awt.ComposePanel
+import org.jetbrains.jewel.bridge.actionSystem.ComponentDataProviderBridge
+import org.jetbrains.jewel.bridge.theme.SwingBridgeTheme
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import javax.swing.JComponent
+
+public fun JewelComposePanel(
+    content: @Composable ComposePanelScope.() -> Unit,
+): JComponent {
+    return ComposePanel().apply {
+        val scope = object : ComposePanelScope {
+            override val panel: JComponent
+                get() = this@apply
+        }
+        setContent {
+            SwingBridgeTheme {
+                CompositionLocalProvider(LocalComponent provides this@apply) {
+                    ComponentDataProviderBridge(this@apply) {
+                        scope.content()
+                    }
+                }
+            }
+        }
+    }
+}
+
+public interface ComposePanelScope {
+
+    public val panel: JComponent
+}
+
+@ExperimentalJewelApi
+public val LocalComponent: ProvidableCompositionLocal<JComponent> = staticCompositionLocalOf {
+    error("CompositionLocal LocalComponent not provided")
+}
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
@@ -11,28 +11,17 @@ import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import javax.swing.JComponent
 
 public fun JewelComposePanel(
-    content: @Composable ComposePanelScope.() -> Unit,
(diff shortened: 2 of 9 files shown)


ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -222,13 +221,9 @@ internal operator fun TextUnit.plus(delta: Float): TextUnit =
         else -> this
     }
 
-internal fun retrieveDensity(): Density {
+internal fun retrieveIdeaDensity(sourceDensity: Density): Density {
```
devkanro, line 224: The component can't be passed here when used in `SwingBridgeService`, so I moved the density calcing out of `SwingBridgeService`.

But I found we call `retrieveIdeaDensity` in composable function, so I change to use the `LocalDensity.current`.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
```diff
@@ -42,14 +41,12 @@ internal class SwingBridgeService(scope: CoroutineScope) {
         return BridgeThemeData(
             themeDefinition = createBridgeThemeDefinition(),
             componentStyling = createBridgeComponentStyling(themeDefinition),
-            density = retrieveDensity(),
```
devkanro, line 45: We can't calc the density in `SwingBridgeService` context, I move out it

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.awt.ComposePanel
+import org.jetbrains.jewel.bridge.actionSystem.ComponentDataProviderBridge
+import org.jetbrains.jewel.bridge.theme.SwingBridgeTheme
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import javax.swing.JComponent
+
+public fun JewelComposePanel(
```
devkanro, line 13: I found the `com.intellij.platform.compose.JBComposePanel` in platform codes, but I think we need provide it in Jewel, so I create one with default theme, and other settings.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+public interface ComposePanelScope {
+
+    public val panel: JComponent
+}
+
+@ExperimentalJewelApi
+public val LocalComponent: ProvidableCompositionLocal<JComponent> = staticCompositionLocalOf {
```
devkanro, line 39: It copied from `com.intellij.platform.compose.JBComposePanel`.

Maybe it won't be used(we provided it in scope), or we will remove it in next release, so I add `@ExperimentalJewelApi` annotation for it.

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/ResourcePainterProvider.kt
```diff
@@ -84,7 +84,7 @@ public class ResourcePainterProvider(
         currentHintsProvider.hints(basePath)
             .forEach { scope.resolveHint(it) }
 
-        val cacheKey = scope.acceptedHints.hashCode()
+        val cacheKey = scope.acceptedHints.hashCode() * 31 + LocalDensity.current.hashCode()
```
devkanro, line 87: The key of bug #236

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
rock3r: Note that Compose/Jewel-related code will be removed from 23.3 (and added back for 24.1) so `JBComposePanel` might not exist until then anyway. When it does exist, though, we should use the platform one.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
rock3r: I think we can omit it, since we don't use it. I prefer the scope, too

rock3r (approved):

rock3r:
Leaving to @Walingar to ship

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
devkanro: @rock3r I think the platform should not provide `JBComposePanel`, it should be provided by Jewel...

rock3r: It will eventually be a moot distinction :)

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/ResourcePainterProvider.kt
Walingar, line 87: Let's then use `density` local variable here

Walingar: Let's surround it with remember(density, scope.acceptedHints)

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
```diff
@@ -0,0 +1,41 @@
...
+                    ComponentDataProviderBridge(this@apply) {
+                        scope.content()
+                    }
+                }
+            }
+        }
+    }
+}
+
+public interface ComposePanelScope {
```
Walingar, line 33: Why do we need this scope? Let's use LocalComponent composition local and use it for retrieveIdeaDensity

Walingar (changes requested):

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/ResourcePainterProvider.kt
devkanro: Using remember may be more expensive. Calculating hashcode is not an expensive operation, which is why I didn't use remember before.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
devkanro: This is just to make the JewelComposePanel DSL more powerful

Walingar: I'd suggest having either scope either LocalComponent, preferrably LocalComponent

rock3r: I see the reasoning. Hadn't considered that. Let's only keep the composition local then, since we don't need the scope anymore.

Walingar (approved):

PR #240, opened 2023-11-02
Optimize performance of SelectableLazyColumn

Walingar:
I've added some optimizations reducing amount of operations during recomposition. 
Also, key events handling is rewritten, taking previous implementation into account, so nothing should be changed, only performance should be optimized.

The only thing that is changed is `DefaultSelectableLazyColumnKeyActions#handleOnKeyEvent` it was broken in 59ff624cdaec319082a89fbbc5512266ee74b953. Thought about making separate PR, but since I wrote tests here, it should be fixed.


foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
```diff
@@ -115,16 +129,17 @@ internal class SelectableLazyListScopeContainer : SelectableLazyListScope {
         selectable: (index: Int) -> Boolean,
         itemContent: @Composable (SelectableLazyItemScope.(index: Int) -> Unit),
     ) {
-        val selectableKeys: List<SelectableLazyListKey> = List(count) {
-            if (selectable(it)) {
-                Selectable(key(it))
-            } else {
-                NotSelectable(key(it))
+        // TODO: [performance] now the implementation requires O(count) operations but should be done in ~ O(1)
```
Walingar, line 132: We should reconsider API for SelectableLazyColumn, so that `allKeys: List<Key>` shouldn't be used at all. Otherwise we have a performance pit here
@fscarponi @rock3r

rock3r: Agreed 👍  We need an API and perf review for this component, since it's foundational and will be used _everywhere_

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -12,118 +12,109 @@ interface SelectableColumnOnKeyEvent {
...
-                }
+        val initialIndex = state.lastActiveItemIndex ?: return
+        val newSelection = ArrayList<Any>(max(initialIndex, state.selectedKeys.size)).apply {
+            addAll(state.selectedKeys)
+        }
+        for (index in initialIndex - 1 downTo 0) {
+            val key = keys[index]
+            if (key is Selectable) {
+                newSelection.add(key.key)
+                state.lastActiveItemIndex = index
```
rock3r, line 37: Does this need to happen in the loop? Wouldn't setting it to the right value at the end be the same, and avoid a bunch of state invalidation? We can do the same thing we do for newSelection, and keep track of the index in a var during the loop

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -12,118 +12,109 @@ interface SelectableColumnOnKeyEvent {
...
-                }
+        val initialIndex = state.lastActiveItemIndex ?: return
+        val newSelection = ArrayList<Any>(max(keys.size - initialIndex, state.selectedKeys.size)).apply {
+            addAll(state.selectedKeys)
+        }
+        for (index in initialIndex + 1..keys.lastIndex) {
+            val key = keys[index]
+            if (key is Selectable) {
+                newSelection.add(key.key)
                 state.lastActiveItemIndex = index
```
rock3r, line 69: Same

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -132,16 +123,14 @@ interface SelectableColumnOnKeyEvent {
      */
     fun onExtendSelectionWithNextItem(keys: List<SelectableLazyListKey>, state: SelectableLazyListState) {
         // todo we need deselect if we are changing direction
```
rock3r, line 125: @fscarponi is this tracked anywhere?

rock3r (commented):
Only a couple of comments, then good to go. Thanks for looking into this!

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
Walingar: Good point, I also thought about it during implementation, but thought that since it happens syncronously nothing bad will happen, hovewer, the state can be collected async, so better to not emit so many values. Agree

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
Walingar: BTW the way how extend selection works now with selectable lazy column is not the same as selection in trees lists in intellij, we will need to handle it

Walingar: I didn't change behaviour in this PR

rock3r: Yeah all these things should be tracked and addressed asap

rock3r: Leaving this action to @fscarponi to tackle on Monday

rock3r (approved):

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
rock3r: Ended up tracking in #242

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyListScope.kt
rock3r: Tracked in #242

Issue #242, opened 2023-11-03
SelectableLazyColumn perf/code review


rock3r:
The `SelectableLazyColumn` is a foundational component that will be used everywhere. As such, we need to make it work as well as it possibly can. This issue is to track the need to do a thorough code and perf review of the component, fixing any issues we find.

Walingar:
Also, there is a usecase that should be taken into account:
> Imagine we have a textfield and lazy column where elements can be selected (e.g. Search Everywhere or Git Branches popup)
I want to propagate arrowDown/arrowUp presses to lazy list. So that when TextField is focused and arrowDown is pressed, lazy list become focused and arrowDown is handled by lazy list.

rock3r:
Related: #470

PR #244, opened 2023-11-03
Fix error appearing when selectedKeys contain invalid key

Walingar:
Selected item keys for items that don't exist anymore don't get purged when the backing data changes, causing stale keys to remain in the map. Pending a more comprehensive cleanup as detailed in #242, this is a work-around that is good enough for now. It is not a long-term solution, to be clear.


rock3r (approved):
LGTM!

PR #251, opened 2023-11-21
LazyTree - contiguous selection appearance

fscarponi:
`LazyTree` selection appearance for contiguous blocks has been fixed.

to achieve this, we check if the previous item key is `IN` the selected keys list.
To reduce the overhead, the selection Key collection has been changed from a `List` to a `Set` (default implementation is LinkedHashSet)

### DIFF
--- a/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
+++ b/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
@@ -18,7 +18,7 @@ public interface SelectableColumnOnKeyEvent {
         for (index in allKeys.indices) {
             val key = allKeys[index]
             if (key is Selectable) {
-                state.selectedKeys = listOf(key.key)
+                state.selectedKeys = setOf(key.key)
                 state.lastActiveItemIndex = index
                 return
             }
@@ -34,7 +34,7 @@ public interface SelectableColumnOnKeyEvent {
         state: SelectableLazyListState,
     ) {
         val initialIndex = state.lastActiveItemIndex ?: return
-        val newSelection = ArrayList<Any>(max(initialIndex, state.selectedKeys.size)).apply {
+        val newSelection = HashSet<Any>(max(initialIndex, state.selectedKeys.size)).apply {
             addAll(state.selectedKeys)
         }
         var lastActiveItemIndex = initialIndex
@@ -59,7 +59,7 @@ public interface SelectableColumnOnKeyEvent {
         for (index in keys.lastIndex downTo 0) {
             val key = keys[index]
             if (key is Selectable) {
-                state.selectedKeys = listOf(key.key)
+                state.selectedKeys = setOf(key.key)
                 state.lastActiveItemIndex = index
                 return
             }
@@ -75,7 +75,7 @@ public interface SelectableColumnOnKeyEvent {
         state: SelectableLazyListState,
     ) {
         val initialIndex = state.lastActiveItemIndex ?: return
-        val newSelection = ArrayList<Any>(max(keys.size - initialIndex, state.selectedKeys.size)).apply {
+        val newSelection = HashSet<Any>(max(keys.size - initialIndex, state.selectedKeys.size)).apply {
             addAll(state.selectedKeys)
         }
         var lastActiveItemIndex = initialIndex
@@ -101,7 +101,7 @@ public interface SelectableColumnOnKeyEvent {
         for (index in initialIndex - 1 downTo 0) {
             val key = keys[index]
             if (key is Selectable) {
-                state.selectedKeys = listOf(key.key)
+                state.selectedKeys = setOf(key.key)
                 state.lastActiveItemIndex = index
                 return
             }
@@ -138,7 +138,7 @@ public interface SelectableColumnOnKeyEvent {
         for (index in initialIndex + 1..keys.lastIndex) {
             val key = keys[index]
             if (key is Selectable) {
-                state.selectedKeys = listOf(key.key)
+                state.selectedKeys = setOf(key.key)
                 state.lastActiveItemIndex = index
                 return
             }
@@ -173,7 +173,7 @@ public interface SelectableColumnOnKeyEvent {
     ) {
         val visibleSize = state.layoutInfo.visibleItemsInfo.size
         val targetIndex = max((state.lastActiveItemIndex ?: 0) - visibleSize, 0)
-        state.selectedKeys = listOf(keys[targetIndex].key)
+        state.selectedKeys = setOf(keys[targetIndex].key)
         state.lastActiveItemIndex = targetIndex
     }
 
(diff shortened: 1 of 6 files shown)


foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -34,7 +34,7 @@ public interface SelectableColumnOnKeyEvent {
         state: SelectableLazyListState,
     ) {
         val initialIndex = state.lastActiveItemIndex ?: return
-        val newSelection = ArrayList<Any>(max(initialIndex, state.selectedKeys.size)).apply {
+        val newSelection = LinkedHashSet<Any>(max(initialIndex, state.selectedKeys.size)).apply {
```
rock3r, line 37: Do you need this to be a linked hash set specifically? That is, do you care about the order?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -206,7 +206,7 @@ public interface SelectableColumnOnKeyEvent {
     ) {
         val visibleSize = state.layoutInfo.visibleItemsInfo.size
         val targetIndex = min((state.lastActiveItemIndex ?: 0) + visibleSize, keys.lastIndex)
-        state.selectedKeys = listOf(keys[targetIndex].key)
+        state.selectedKeys = setOf(keys[targetIndex].key)
```
rock3r, line 209: This seems incoherent with the rest, why is this not a linked hash set?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BasicLazyTree.kt
```diff
@@ -150,8 +150,19 @@ public fun <T> BasicLazyTree(
                         ?: false,
                 )
 
-            val backgroundShape by remember {
-                mutableStateOf(RoundedCornerShape(elementBackgroundCornerSize))
+            val backgroundShape by derivedStateOf {
+                val upperCorners = flattenedTree.getOrNull(index - 1)?.id?.let {
```
rock3r, line 154: I would name these two flags as `hasRoundedTopCorners` and `hasRoundedBottomCorners`

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BasicLazyTree.kt
```diff
@@ -150,8 +150,19 @@ public fun <T> BasicLazyTree(
...
+                    it !in treeState.delegate.selectedKeys
+                } ?: false
+                val bottomCorners = flattenedTree.getOrNull(index + 1)?.id?.let {
+                    it !in treeState.delegate.selectedKeys
+                } ?: false
+                RoundedCornerShape(
+                    topStart = if (upperCorners) elementBackgroundCornerSize else CornerSize(0.dp),
```
rock3r, line 161: This if-else is repeated 4 times, I'd extract a small function for this. Something like:

```kotlin
private fun computeCornerSize(roundedSize: CornerSize, rounded: Boolean) =
    if (rounded) roundedSize else CornerSize(0.dp)
```

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/DefaultTreeViewOnKeyEvent.kt
```diff
@@ -41,7 +41,7 @@ public open class DefaultTreeViewOnKeyEvent(
                             ?.let {
                                 state.lastActiveItemIndex =
                                     keys.indexOfFirst { selectableKey -> selectableKey.key == parentNodeKey }
-                                state.selectedKeys = listOf(parentNodeKey)
+                                state.selectedKeys = setOf(parentNodeKey)
```
rock3r, line 44: This and the next few usages also look dodgy, why not linked here?

rock3r (changes requested):

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
fscarponi: it is by default under the hood

fscarponi: but it is ok to move it to a generic Set everywhere

rock3r: I'd rather use a generic Set unless you specifically need it to be linked. The implementation under the hood doesn't matter, and using the simple Set instead communicates different expectations.

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
fscarponi: we cannot move to generic Set, it does not have the required constructor, and I will not change the performance review made by @Walingar

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/DefaultTreeViewOnKeyEvent.kt
fscarponi: we don't need here a more performant constructor

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
rock3r: I'm sorry, I don't follow. Set is an interface, what constructor are you referring to? And what's the perf implication of using a non-linked hash set instead of a linked one?

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/DefaultTreeViewOnKeyEvent.kt
rock3r: I don't understand what this means

fscarponi: nvm, probably I misunderstood

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -240,7 +240,7 @@ public interface SelectableColumnOnKeyEvent {
      * Select All.
      */
     public fun onSelectAll(keys: List<SelectableLazyListKey>, state: SelectableLazyListState) {
-        state.selectedKeys = keys.filterIsInstance<Selectable>().map { it.key }
+        state.selectedKeys = keys.filterIsInstance<Selectable>().map { it.key }.toCollection(HashSet())
```
rock3r, line 243: Why not `toSet()`?

rock3r (approved):

PR #257, opened 2023-11-27
Remove logging

### LINKED ISSUE #256: Remove logging from foundation and UI
We shouldn't have logging in the foundation and UI modules. This issue is to remove it.

rock3r:
Fixes #256


devkanro (approved):
When necessary, you can use debugMode to print logs

rock3r:
Yes, but it shouldn't be necessary to begin with at this point. Debug logging can be added if/when needed while debugging.

PR #262, opened 2023-12-01
Fix scrollbar color when have 0f alpha

### LINKED ISSUE #259: Scrollbar in light theme have wrong color
hovered colour is ok, but the default is not. In dark mode, it is correct!
image
image

fscarponi:
Fixes #259


rock3r (approved):
It's very sad we have to resort to these hacks, but nothing we can do about it. I am still not sure why this key, that was never present until the last EAP build, now is but is set to transparent.

PR #276, opened 2023-12-18
Extend TabStrip API to support custom behaviours

fscarponi:
In the IDE and in some specs, you can find some kind of fanciful Tabs usage...

as an example, using a composite tabHeader content that has a title made by two different texts with different styles or icons

image

image

we can change the api for to allow users to provide content for tabs, allowing users to solve their own problems

migrating to a new DSL:

```
    public class TabData.Editor(
        override val selected: Boolean,
        override val content: @Composable TabContentScope.(tabState: TabState) -> Unit,
        override val closable: Boolean = true,
        override val onClose: () -> Unit = {},
        override val onClick: () -> Unit = {},
    ) : TabData()
```

and exposing the following methods:

```
@Composable
public fun TabContentScope.SimpleTabContent(
    title: String,
    state: TabState,
    icon: Painter?,
) 

@Composable
public fun TabContentScope.SimpleTabContent(
    state: TabState,
    icon: (@Composable () -> Unit)? = null,
    label: @Composable () -> Unit,
)
```

this will allow achieving custom behavior for the tab, as we show in the standalone sample
image

### DIFF
--- a/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Tabs.kt
+++ b/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Tabs.kt
@@ -2,6 +2,7 @@
 
 package org.jetbrains.jewel.samples.standalone.view.component
 
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -14,16 +15,21 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.drawWithCache
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.dp
 import org.jetbrains.jewel.foundation.theme.JewelTheme
 import org.jetbrains.jewel.samples.standalone.StandaloneSampleIcons
 import org.jetbrains.jewel.samples.standalone.viewmodel.View
 import org.jetbrains.jewel.ui.component.Icon
 import org.jetbrains.jewel.ui.component.IconButton
+import org.jetbrains.jewel.ui.component.SimpleTabContent
 import org.jetbrains.jewel.ui.component.TabData
 import org.jetbrains.jewel.ui.component.TabStrip
 import org.jetbrains.jewel.ui.component.Text
+import org.jetbrains.jewel.ui.painter.rememberResourcePainterProvider
 import org.jetbrains.jewel.ui.theme.defaultTabStyle
+import org.jetbrains.jewel.ui.util.thenIf
 import kotlin.math.max
 
 @Composable
@@ -48,7 +54,16 @@ private fun DefaultTabShowcase() {
         tabIds.mapIndexed { index, id ->
             TabData.Default(
                 selected = index == selectedTabIndex,
-                content = { Text(text = "Default tab $id") },
+                content = {
+                    val iconProvider =
+                        rememberResourcePainterProvider("icons/search.svg", StandaloneSampleIcons::class.java)
+                    val icon by iconProvider.getPainter()
+                    SimpleTabContent(
+                        state = it,
+                        title = "Default Tab $id",
+                        icon = icon,
+                    )
+                },
                 onClose = {
                     tabIds = tabIds.toMutableList().apply { removeAt(index) }
                     if (selectedTabIndex >= index) {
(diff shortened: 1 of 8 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
```diff
@@ -56,9 +57,10 @@ public class DefaultComponentStyling(
     public val circularProgressStyle: CircularProgressStyle,
     public val defaultButtonStyle: ButtonStyle,
     public val defaultDropdownStyle: DropdownStyle,
-    public val defaultTabStyle: TabStyle,
-    public val dividerStyle: DividerStyle,
+    public val toolWindowTabStyle: TabStyle,
```
rock3r, line 60: Please keep the alphabetical sorting of the parameters

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -81,28 +82,44 @@ public fun TabStrip(
...
-        override val onClose: () -> Unit = {},
-        override val onClick: () -> Unit = {},
-    ) : TabData()
+    public sealed class Default : TabData() {
+
+        public abstract val label: String
+
+        @Immutable
+        @GenerateDataFunctions
+        public class ToolWindowTab(
```
rock3r, line 97: Either this is `Toolwindow`, or the others are `EditorTab` and `CustomTab`, for consistency. I'd much rather the first. I am also not sure why you renamed `Default` to `Toolwindow`, aren't there any other tab implementations that use that style? I doubt it...

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -81,28 +82,44 @@ public fun TabStrip(
...
+            override val closable: Boolean = true,
+            override val onClose: () -> Unit = {},
+            override val onClick: () -> Unit = {},
+        ) : Default()
+    }
 
     @Immutable
     @GenerateDataFunctions
-    public class Editor(
+    public class Custom(
```
rock3r, line 120: I disagree with this change. I think that "Custom" is not a real kind of tab. There are "regular" tabs (also known as `Default`), and `Editor` tabs. The difference between the two is the styling. Custom tabs are... what? Editor tabs? Non-editor tabs? Being custom is not a _style_ difference, it's a _contents_ difference.

I believe it would be much more productive to have the two variants from before (`Default` and `Editor`) but instead of carrying a `label` parameter, they carry a `content` parameter that is a composable lambda. That's the only change needed to implement what you want to achieve.

If you for whatever reason want to also keep the label-based DSL, you can easily implement it by calling the content-based API with a `Text(label)`.

rock3r (changes requested):

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -47,19 +48,16 @@ public fun TabStrip(
...
+            modifier = Modifier.horizontalScroll(scrollState).scrollable(
+                orientation = Orientation.Vertical,
+                reverseDirection = ScrollableDefaults.reverseDirection(
+                    LocalLayoutDirection.current,
+                    Orientation.Vertical,
+                    false,
+                ),
+                state = scrollState,
+                interactionSource = remember { MutableInteractionSource() },
+            ).selectableGroup(),
```
rock3r, line 60:
```suggestion
            modifier = Modifier.horizontalScroll(scrollState)
                .scrollable(
                    orientation = Orientation.Vertical,
                    reverseDirection = ScrollableDefaults.reverseDirection(
                        LocalLayoutDirection.current,
                        Orientation.Vertical,
                        false,
                    ),
                    state = scrollState,
                    interactionSource = remember { MutableInteractionSource() },
                )
                .selectableGroup(),
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -91,23 +89,71 @@ public sealed class TabData {
...
         override val onClose: () -> Unit = {},
         override val onClick: () -> Unit = {},
-    ) : TabData()
+    ) : TabData() {
+
+        @Deprecated(
+            "Use the primary constructor instead",
+            ReplaceWith(
+                "Editor(selected = selected, content = { Text(text) }, icon = icon, closable = closable, onClose = onClose, onClick = onClick)",
+                "org.jetbrains.jewel.ui.component.TabData.Default",
```
rock3r, line 138:
```suggestion
                "org.jetbrains.jewel.ui.component.TabData.Editor",
```

Re-read when you copy paste, please 😁

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tabs.kt
```diff
@@ -127,11 +125,10 @@ internal fun TabImpl(
                 Image(modifier = Modifier.alpha(iconAlpha), painter = icon, contentDescription = null)
             }
 
-            Text(
-                modifier = Modifier.alpha(labelAlpha),
-                text = tabData.label,
-                color = tabStyle.colors.contentFor(tabState).value,
-            )
+            Box(Modifier.alpha(labelAlpha)) {
```
rock3r, line 128: Shouldn't `labelAlpha` be called `contentAlpha` if it affects the entirety of the content? And, are you sure this alpha affects all the content and not just the label?

rock3r (changes requested):

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tabs.kt
fscarponi: ok for content alpha naming...

about expected Behaviour. Who knows?
as example, i have a friend in cwm on the "Slider.kt" tab... but i don't see the icon! It appears (sometimes) when i hover the element with mouse
image
or if he moves between tabs, an empty box appears sometimes, but not always....
image

in another following case he is in a file, and the icon does not appear either that file is open and the editor is active...
image

with this PR you can achieve this behavior... but spec does not mention how it should look...

my point is that we can hold it "as is" and monitor if it fits futures use cases...

rock3r: Yeah, fair enough. The Swing implementations are broken enough to begin with that it's almost impossible to figure out what the visual intention is. We can revisit later.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Tabs.kt
```diff
@@ -83,7 +84,7 @@ private fun EditorTabShowcase() {
         tabIds.mapIndexed { index, id ->
             TabData.Editor(
```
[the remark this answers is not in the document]
rock3r: Doesn't look resolved (commit not pushed?)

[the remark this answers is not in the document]
fscarponi: yea

rock3r (approved):

PR #280, opened 2023-12-28
README Update - Remove "platform-specific" from dependency for plugin

Chozzle:
This took me a few hours to debug. Using "platform-specific" does not load some common files. For example the code in the sample won't compile because `org.jetbrains.jewel.bridge.addComposeTab` is missing.

Maybe the code published was changed, but the README not updated?


rock3r:
Hi @Chozzle thanks for the readme fixup! I think the "platform specific" bit in the readme was intended to be a placeholder, but it's now clearer with this PR. I'll do a review and check the code in the readme is working, thanks for the heads up

rock3r (approved):

PR #289, opened 2024-01-27
Allow cross builds via construo plugin

fourlastor:
Hello! I'm the author of [construo](https://github.com/fourlastor-alexandria/construo), a plugin to cross compile JVM apps. This draft is to give an example on how it could be used to build jewel for different OSes


rock3r:
Hi Daniele, thanks for the PR! Let me know when the Windows issue is solved :)

samples/standalone/build.gradle.kts
```diff
@@ -1,10 +1,13 @@
 @file:Suppress("UnstableApiUsage")
 
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
+import io.github.fourlastor.construo.Target
 
 plugins {
     jewel
     alias(libs.plugins.composeDesktop)
+    alias(libs.plugins.construo)
+    alias(libs.plugins.shadow)
```
rock3r, line 10: I don't see where this is used?

samples/standalone/build.gradle.kts
```diff
@@ -52,3 +55,42 @@ tasks {
...
+    }
+    jlink {
+        // required as it's used via string by compose, so it's not picked up by jlink automatically
+        modules.addAll("jdk.zipfs")
+    }
+    targets {
+        // Linux X64
+        create<Target.Linux>("linuxX64") {
+            architecture.set(Target.Architecture.X86_64)
+            jdkUrl.set("https://cache-redirector.jetbrains.com/intellij-jbr/jbr-17.0.10-linux-x86-b1087.17.tar.gz")
```
rock3r, line 80: Would this be possible to automate using the foojay.io APIs (i.e., the list of [distributions](https://api.foojay.io/disco/v3.0/distributions) and releases)? You can use the same logic as the [FooJay Gradle plugin](https://github.com/gradle/foojay-toolchains/blob/main/foojay-resolver/src/main/kotlin/org/gradle/toolchains/foojay/FoojayApi.kt#L19) does to then formulate a list of available distributions matching a `JavaToolchainRequest` and have Gradle download the right one. [This](https://github.com/gradle/foojay-toolchains/blob/main/foojay-resolver/src/main/kotlin/org/gradle/toolchains/foojay/FoojayApi.kt#L34) is a good entry point, I suppose. You might even want to just depend on the FooJay plugin and reuse their logic.

samples/standalone/build.gradle.kts
fourlastor: Shadow is used to produce a far jar, ie a jar which includes all the dependencies as well (which is needed to ridistribute it). Compose does a similar thing behind the scenes, just not using shadow

fourlastor:
roast is no longer considered a virus :tada: :grinning: from MS:

>  Open command prompt as administrator and change directory to c:\Program Files\Windows Defender
        Run “MpCmdRun.exe -removedefinitions -dynamicsignatures”
        Run "MpCmdRun.exe -SignatureUpdate"

I tested (and had the original reporter testing it as well) and it could run without issues :sparkles:

samples/standalone/build.gradle.kts
rock3r: @fourlastor thoughts on this? Is it even doable?

fourlastor: I haven't had the chance to check it out yet, it would be pretty nice if it was possible!

fourlastor:
Sure, a note though that construo doesn't support signing the executable yet, which is why I opened this as a draft

rock3r:
I think it's fine if it's well documented, and you reckon you'll add the signing to it and come back to update

fourlastor:
Hi! Sorry for the long wait.

I can bring this to a shippable state, but without signing. We could still sign with apple's services (running the ci on a macos runner), but I doubt I'll be able to finish the signing part of construo anytime soon, both because I'm very unfamiliar with the process and because I physically don't have a mac to test the thing out :sweat_smile: 

There's been a few major changes in construo since, the most important being that I removed the integration with FooJay, as that more often than not leaded to the wrong JDK url being selected, I don't plan on adding it back as I think having a fixed url for the distribution JDK which is updated manually is a much safer and predictable process.

Let me know if this is still something that would work for jewel, and in that case I'll update the PR right away!

rock3r:
@fourlastor yeah that still sounds pretty valuable. As long as the readme mentions that signing is up to you, I think it's fine :)

docs/distribute.md
```diff
@@ -0,0 +1,9 @@
+# Creating a self contained application
+
+You can package the standalone sample with the following gradle tasks:
+
+- Linux `samples:standalone:packageLinuxX64`
+- Macos `samples:standalone:packageMacOsM1`
```
rock3r, line 6: Nit:

```suggestion
- macOS `samples:standalone:packageMacOsM1`
```

I'd also rename the task to `packageMacOsArm64` or `packageMacOsAarch64` but not sure if it's a Construo thing :)

gradle/libs.versions.toml
```diff
@@ -1,19 +1,21 @@
...
+intelliJPlatformBuild = "243.21155.17-EAP-SNAPSHOT"
+ideaPlugin = "2.1.0"
 jna = "5.14.0"
 kotlin = "2.0.21-RC"
 kotlinSarif = "0.5.0"
-kotlinpoet = "1.15.2"
-kotlinterGradlePlugin = "3.16.0"
-kotlinxSerialization = "1.5.1"
-kotlinxBinaryCompat = "0.14.0"
-poko = "0.13.1"
```
rock3r: I am very confused by this diff... did you rebase your branch on main?

fourlastor: I merged on main, but I think you're looking at the merge commit (if I look at the "files" tab I only see construo and shadow being added)

rock3r: Yeah it was GH that for some reason had chosen to show me a subset of commits

rock3r:
@fourlastor is the task renaming doable? If it's a Construo thing, then we'll do it separately :) and ship this

docs/distribute.md
fourlastor: [db7860f](https://github.com/JetBrains/jewel/pull/289/commits/db7860fbcc457bbd882430f762006ad346658c97) it's definitely possible, the `package*` task depend on the name of the target, which is just a string. So if you created a target with `create<Target.MacOs>("sebastiano")`, you'd have a `packageSebastiano` task :D

Updated the `DISTRIBUTING.md` file and put it in the root, and updated the target name to be `macOsAarch64`, so now the packaging task will be called `packageMacOsAarch64`

fourlastor: [08e5db8](https://github.com/JetBrains/jewel/pull/289/commits/08e5db8ad6a1ecceabc6c28716ac3a7e4819ada5) forgot the nit

gradle/libs.versions.toml
```diff
@@ -55,3 +57,4 @@ kotlinx-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", vers
 kotlinter = { id = "org.jmailen.kotlinter", version.ref = "kotlinterGradlePlugin" }
 ktfmt = { id = "com.ncorti.ktfmt.gradle", version.ref = "ktfmtGradlePlugin" }
 poko = { id = "dev.drewhamilton.poko", version.ref = "poko" }
+shadow = { id = "com.github.johnrengelman.shadow", version.ref = "shadow"}
```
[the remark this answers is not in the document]
fourlastor: [b55c4da](https://github.com/JetBrains/jewel/pull/289/commits/b55c4daf31b5f64815ae192d91e9d0a168014f48)

rock3r (approved):

crowforkotlin:
> @fourlastor is the task renaming doable? If it's a Construo thing, then we'll do it separately :) and ship this

https://github.com/JetBrains/JetBrainsRuntime/issues/477

I think Construo actually just needs to fix the problem on the Mac, but that's been implemented so far. Moreover, JBR for Windows and Linux comes with JBoy, so it can directly coexist with Jewel and package it. But jpackage is missing on the mac, I don't know why, I'll try to use construo first to see if that solves my problem

rock3r:
Closing as this has been ported to the `intellij-community` repo

PR #290, opened 2024-01-30
Add support for IJP 24.1

rock3r:
This PR implements initial support for the IntelliJ Platform 24.1. It also:
 * Bumps the 233 dependency to 233.3
 * Adds setup instructions to the README
 * Updates a number of dependencies
 * Cleans up the build scripts
 * Adds 241 tasks to the CI
 * Builds the standalone themes based on 241

To note, running the ide-plugin sample as-is in 241 will crash. This is because Jewel is bundled in 241, but it's an older version, and that breaks in all sorts of fun ways. The only known workaround for now is shadowing everything.


fscarponi (approved):
I am quite confident that the PR can be shipped, I would suggest waiting for the release of 241 stable. Later if we don't need to create a new bridge to compensate for the differences, I would reconsider the bridge naming (to help the user with a more intuitive approach)

rock3r:
@fscarponi what exactly do you think is a blocker that requires waiting for 241 stable in this PR? There isn't any new logic or 241-specific code in Jewel, it's mostly about the build

rock3r:
Note: talked online about this and the worries are mostly about 241 changing LaF keys and values again, but this still seems to work fine for now, at least in the IDE.

lamba92 (approved):

rock3r:
There are some breakages in standalone due to significative changes in how SVGs are patched in 241 — I will address them in a separate PR.

Issue #291, opened 2024-01-31
Update SVG patching for new logic in 241


rock3r:
In IJP 241, as of [IDEA-342269](https://youtrack.jetbrains.com/issue/IDEA-342269/Colorizing-CheckBoxes-RadioButtons-via-theme-doesnt-work-in-new-UI), the logic to patch colors checkbox and radio button SVGs has changed. I am still trying to figure out the exact logic, but this requires at a minimum:

* an update of the actual SVG files (which are at least now less wonky than they were!)
* an update of the patching logic
* an update of the validation outline size/offset to match the finally correct SVGs
* confirming everything works

The first two should only affect standalone, since we use the patching as provided by the IDE for the bridge (always with the hope it's not terribly broken, which it maybe still is).

rock3r:
Fixed in #293

PR #293, opened 2024-02-01
#291 — Fix icons patching and theming issues in 241

rock3r:
This PR fixes the SVG icons patching for checkboxes and radio buttons in 241 and on standalone, by:
1. Using the SVGs from 241 in standalone, which are 10000% less wonky than the 23* ones
2. Changing the patching logic for their colours to use the new ID-based one introduced in 241
3. Fixing checkbox and radio button implementations to remove hacks and reliably work with the new logic/SVGs
4. Resolve errors in bridge by following the hard deprecation notices' instructions

 Before | After
 --- | ---
  | 
  | 

 Before | After
 --- | ---
  | 
  | 
  | 

This closes #290


rock3r:
EAP 3 is just out so I'm double checking everything just in case :)

rock3r:
One thing left to fix: checkboxes/radio buttons in classic UI themes (Darcula, High Contrast) don't get colours properly patched

rock3r:
Verified to work on both standalone and IDE. Required fixes to the painter hints order, a buncha extra checks (e.g. only applying the new colour patching scope on New UI themes — that are, for now, hardcoded), and some other fun acrobatics.

I took the opportunity to add KDocs to a bunch of things and to test on Windows, too. I had to fix a few issues with bloody \r\n's and / vs \, of course.

fscarponi (approved):

PR #303, opened 2024-02-17
Fix remembers painter twice and break the cache

devkanro:
1. Move the InDebugMode property to foundation module.
2. Rename the utils in foundation to util
3. Fix remembers painter twice to break the cache, see the comments below.


ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/ResourcePainterProvider.kt
```diff
@@ -166,7 +166,7 @@ public class ResourcePainterProvider(
                     loadSvgPainter(inputStream, scope)
                 }
             },
-            rememberAction = { remember(url, scope.density) { it } },
+            paintAction = { it },
```
devkanro, line 169: We should not remember painter here, because we have already the cache here.

Before:
截屏2024-02-17 19 30 59

After:
截屏2024-02-17 19 31 57

It only happened with first default selected icon button

ui/src/main/kotlin/org/jetbrains/jewel/ui/painter/ResourcePainterProvider.kt
```diff
@@ -218,14 +218,14 @@ public class ResourcePainterProvider(
...
-        rememberAction = { remember(url, scope.density) { it } },
+        paintAction = { it },
     )
 
     @Composable
     private fun <T> tryLoadingResource(
         url: URL,
         loadingAction: (URL) -> T,
-        rememberAction: @Composable (T) -> Painter,
+        paintAction: @Composable (T) -> Painter,
```
devkanro, line 228: So I rename the rememberAction to paintAction, it more seems like a action to convert T to a painter

rock3r (approved):
Thanks!

PR #310, opened 2024-02-22
Fix cpu usage from TexField caret animation by updating compose dependencies

fscarponi:
Refer to CMP: TextField cursor animation uses a lot of CPU [#4225](https://github.com/JetBrains/compose-multiplatform/issues/4225)


samples/ide-plugin/build.gradle.kts
```diff
@@ -18,6 +18,8 @@ repositories {
     maven("https://www.jetbrains.com/intellij-repository/releases")
     maven("https://cache-redirector.jetbrains.com/intellij-dependencies")
     mavenCentral()
+    google()
```
rock3r, line 21: Please put it above Central, so it's looked at first

rock3r (changes requested):

rock3r (approved):

PR #327, opened 2024-03-22
Load markdown images with Coil3

obask:


markdown/README.md
```diff
@@ -12,6 +12,8 @@ Additional supported Markdown, via extensions:
 
 * Alerts ([GitHub Flavored Markdown][alerts-specs]) — see [`extension-gfm-alerts`](extension-gfm-alerts)
 
+Im**agetime, and no guarantees are made for binary and source compatibility. It might also have bugs and missing features. ![here is an image](https://hellorse.fr/images/p/media/image/products/product_76409_2b3beddbf246765581c5e91a9fd6.webp?p=product_thumbnail&type=product) And ot**her stuff time, and no guarantees are made for binary and source compatibility. It might also have bugs and missing features. time, and no guarantees are made for binary and source compatibility. It might also have bugs and missing features.
```
rock3r, line 15: This change looks wrong

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -144,7 +151,7 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
     private fun Image.toMarkdownImageOrNull(): MarkdownBlock.Image? {
         if (destination.isBlank()) return null
 
-        return MarkdownBlock.Image(destination.trim(), title.trim())
+        return MarkdownBlock.Image(destination.trim(), (title?:"").trim())
```
rock3r, line 154:
```suggestion
        return MarkdownBlock.Image(destination.trim(), title.orEmpty().trim())
```

markdown/core/build.gradle.kts
```diff
@@ -8,6 +8,10 @@ plugins {
 dependencies {
     compileOnly(projects.ui)
     api(libs.commonmark.core)
+    implementation("io.coil-kt.coil3:coil-compose:3.0.0-alpha06")
+//    Exception in thread "AWT-EventQueue-0" java.lang.IllegalStateException: Module with the Main dispatcher is missing. Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android' and ensure it has the same version as 'kotlinx-coroutines-core'
+//	https://github.com/coil-kt/coil/issues/842#issuecomment-1975115060
+    runtimeOnly("org.jetbrains.kotlinx:kotlinx-coroutines-swing:1.8.1-Beta")
```
rock3r, line 14: This will break in the IDE, we can't bring coroutine dependencies in. We need to use those that the platform has.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -95,10 +95,17 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
         when (this) {
             is BlockQuote -> toMarkdownBlockQuote()
             is Heading -> toMarkdownHeadingOrNull()
-            is Paragraph -> toMarkdownParagraphOrNull()
+            is Paragraph -> {
+                val child = firstChild
+                if (child is Image && child === lastChild) {
```
fscarponi, line 100: @rock3r Could this be written better? (does "===" operator be error prone?)

rock3r: The `===` check makes the `is` check pointless I suppose — but I haven't checked the code too thoroughly yet, since Oleg said it's not really working anyway. I'm not sure why we need a ref equality check here.

obask: `is` check is needed because it should be a paragraph with one chunk of text

markdown/core/build.gradle.kts
```diff
@@ -9,6 +9,10 @@ plugins {
 dependencies {
     api(projects.ui)
     api(libs.commonmark.core)
+    api(libs.ktor.client.java)
```
rock3r, line 12: Why is Ktor exposed as public API by the core module? I don't think the core module should expose it, or any networking API, to users.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -131,3 +137,17 @@ public fun LazyMarkdown(
...
 }
+
+private const val IMAGES_MEMORY_CACHE_SIZE = 24L * 1024 * 1024 // 24mb
+
+/**
+ * This method sets up an image loader with a memory cache but disables the disk cache. Disabling the disk cache is
+ * necessary because Coil crashes when attempting to use the file system cache with IDEA platform.
+ *
+ * Otherwise Coil3 will throw java.lang.NoSuchMethodError: kotlinx.coroutines.CoroutineDispatcher.limitedParallelism
+ */
```
rock3r, line 148: Have you filed a Coil bug for this? Or is there a workaround, like moving the init to the bridge and standalone modules, handling things differently in each case? For what you write, disk caching should work fine in standalone as-is, and may require a different approach for the bridge, where we provide a custom implementation that works in IJP.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
rock3r: That check is for `Image`, not for text... as far as I can tell, this is checking if this is just a paragraph block that only contains an image, in which case we render it as block-level image by ignoring the paragraph wrapper. Is that correct?

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -89,6 +101,7 @@ public open class DefaultMarkdownBlockRenderer(
             is IndentedCodeBlock -> render(block, rootStyling.code.indented)
             is Heading -> render(block, rootStyling.heading, enabled, onUrlClick, onTextClick)
             is HtmlBlock -> render(block, rootStyling.htmlBlock)
+            is MarkdownBlock.Image -> render(block, rootStyling.image)
```
rock3r, line 104: Nit: import Image for consistency with other cases

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -103,6 +116,8 @@ public open class DefaultMarkdownBlockRenderer(
         }
     }
 
+    private data class ImageSize(val w: TextUnit, val h: TextUnit)
```
rock3r, line 119:
```suggestion
    private data class ImageSize(val width: TextUnit, val height: TextUnit)
```

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -125,9 +141,27 @@ public open class DefaultMarkdownBlockRenderer(
                     .clickable(interactionSource = interactionSource, indication = null, onClick = onTextClick),
             text = renderedContent,
             style = mergedStyle,
+            inlineContent = renderedImages(block.inlineContent, knownSizes),
         )
     }
 
+    private fun getImages(input: List<InlineMarkdown>) =
+        buildList<InlineMarkdown.Image> {
+            fun dfs(items: List<InlineMarkdown>) {
```
rock3r, line 150: I am not sure what `dfs` means. Depth-first search? Please use more descriptive names

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -434,6 +473,37 @@ public open class DefaultMarkdownBlockRenderer(
...
+                width = knownSizes[image.source]?.w ?: 100.sp,
+                height = knownSizes[image.source]?.h ?: 100.sp,
+                placeholderVerticalAlign = PlaceholderVerticalAlign.Bottom,
+            )
+        ) {
+            AsyncImage(
+                model =
+                    ImageRequest.Builder(LocalPlatformContext.current).data(image.source).size(Size.ORIGINAL).build(),
+                contentDescription = image.title,
+                // TODO: figure out what to use instead of *.sp
```
rock3r, line 499: Why are you not using dp?

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
obask: Yes. Block images are rendered in one pass while for inline images we have to remember size and update a placeholder, so I wanted to process them separately. E.g. block images are just images, they are more common use case and  less likely to cause problems.

markdown/core/build.gradle.kts
obask: It should actually be `runtimeOnly`. Changed. It took me a while to figure out why IDEA plugin's ktor results in "io.ktor.client.HttpClientEngineContainer: io.ktor.client.engine.java.JavaHttpEngineContainer not a subtype" erorr, but adding this dependency before ktor somehow resolves it.

rock3r: Still reads api, did you push?

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
rock3r: >This workaround is not needed for standalone mode.

So it should only be done in the bridge version, and indeed we need an issue to track checking this again when we move to the IJP

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
obask: 💯

markdown/core/build.gradle.kts
obask: not yet, sorry. Github is confusing. It wasn't clear if I can collect all comment and request re-review or resolve them one by one.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
obask: `Markdown` is invoked directly from the example code. Do you know how to add it to the bridge only?
Additionally, if we change IDE vs Standalone behaviour, it would be much more difficult to figure out if something is wrong in one but works fine in another.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -434,6 +474,51 @@ public open class DefaultMarkdownBlockRenderer(
...
+                // but simply maintains behavior consistent with standalone AsyncImage rendering.
+                val imageSize = knownSizes[image.source]
+                Placeholder(
+                    width = imageSize?.width?.toSp() ?: 20.sp,
+                    height = imageSize?.height?.toSp() ?: 20.sp,
+                    placeholderVerticalAlign = PlaceholderVerticalAlign.Bottom,
+                )
+            }
+        ) {
+            AsyncImage(
```
rock3r, line 505: BTW you need to make sure Ktor is initialised correctly (e.g., in the IDE it needs to respect the IDE proxy settings).

As a starting point, you can do something like this: https://gist.github.com/rock3r/6f633cb0e5b73625cc6ec0570896ff85

Of course this needs more work than just copy pasting it, but it should point you in the right direction

rock3r:
@obask for inline images, you can probably take inspiration from https://github.com/halilozercan/compose-richtext/blob/main/richtext-ui/src/commonMain/kotlin/com/halilibo/richtext/ui/string/InlineContent.kt

obask:
> @obask for inline images, you can probably take inspiration from https://github.com/halilozercan/compose-richtext/blob/main/richtext-ui/src/commonMain/kotlin/com/halilibo/richtext/ui/string/InlineContent.kt

I think it does the same `.toSp()` conversion as a previous version of the code. I also updated the way we handle state and deleted a special case for block images.

obask:
And I fixed images scaling, BTW. It shows them the same way as other markdown editors.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: Be warned you'll have to recreate the branch and PR on the IJ repo. You'll also need to manage both Gradle and JPS builds in parallel

obask:
moved this PR to https://github.com/JetBrains/intellij-community/pull/2924, but still figuring out `*.iml` dependencies

Issue #329, opened 2024-03-23
Consider deleting `intermediateMarkdown` strings in the intermediate representation and use CommonMark `Node` trees instead


obask:
An example with working samples is located in 3df1cccc96211384a767b88a70cfa21cdca3ac9f.
This should improve parsing speed and reduce GC overload. Also, the biggest improvement will be in markdown extensions API since we'll be able to reuse CommonMark patterns and only extend it in for rendering.

Please let me know what you think!

PS of course string rendering code could be later cleaned up too, this is a minimal set of changes making samples to run

rock3r:
It's kinda funny cause as I was writing the initial version I went back and forth several times between the two versions, passing along the Node vs only a string of the inline markdown. I ended up on the string because I didn't think extensions would need to access the underlying Node(s) — if they did, most of the block-level parsing would be duplicated effort. I thought, in that case the proper fix is to make sure all the required info is there in our own block level models. The added benefit is that you can pretty easily swap out Commonmark for another parser if you desire, for example to support other Markdown-like formats such as reText (not that I was planning to).

Do you have perf improvement numbers? Those could be the deciding factor, if the difference is significant enough.

However, I think a bigger perf improvement would come if we somehow managed to support incremental updating, e.g., by tracking with block(s) the user is editing in the text area, and only re-processing those

rock3r:
Oh and one more reason I forgot to mention — block-level models as they are now can easily be made serializable, as they're only made of models we control and primitive types. This might be useful in the future (e.g., if the first parsing step result needs to be stored to disk or sent over network)

obask:
Here are my measurements(this comment is just to preserve benchmark results but I'd like to discuss over a phone about overall approach and if it makes sense at all)...

exact code: https://github.com/obask/jewel/commit/aef1b88c4e562730ffca186d2dd68251f3993067
a large markdown file for benchmarking(this commit was cherry-picked on top): https://github.com/obask/jewel/blob/7c9345fd2b18158073a0319e2477f5be4ac31fea/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/markdown/MarkdownCatalog.kt

Initial measurements for rendering:
-----------------------------
method: DefaultMarkdownBlockRenderer.render
### COMMONMARK:

Call count: 10,000
Total wall time: 363 ms

**Average wall time: 36 μs**
Max wall time: 2,931 μs

MAIN:

Call count: 10,011
[...]

PR #331, opened 2024-03-27
Use value classes for inline markdown #329

obask:
This classes wrap CommonMark's nodes and preserve
a familiar API. Some tests were turned off because we don't use the same IR for string serialization.

Changing dataclasses to values classes for blocks
is coming in the next pull request.

### DIFF
--- a/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
+++ b/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
@@ -34,16 +34,10 @@ import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.markdown.InlineMarkdown
 import org.jetbrains.jewel.markdown.MarkdownBlock
 import org.jetbrains.jewel.markdown.MarkdownBlock.CodeBlock
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H1
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H2
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H3
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H4
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H5
-import org.jetbrains.jewel.markdown.MarkdownBlock.Heading.H6
-import org.jetbrains.jewel.markdown.MarkdownBlock.ListBlock.UnorderedList
 import org.jetbrains.jewel.markdown.MimeType
 import org.jetbrains.jewel.markdown.extensions.MarkdownProcessorExtension
 import org.jetbrains.jewel.markdown.rendering.DefaultInlineMarkdownRenderer
+import org.jetbrains.jewel.markdown.toInlineNode
 
 /**
  * @param optimizeEdits Optional. Indicates whether the processing should only update the changed blocks
@@ -205,7 +199,6 @@ public class MarkdownProcessor(
             is Paragraph -> toMarkdownParagraphOrNull()
             is FencedCodeBlock -> toMarkdownCodeBlockOrNull()
             is IndentedCodeBlock -> toMarkdownCodeBlockOrNull()
-            is Image -> toMarkdownImageOrNull()
             is BulletList -> toMarkdownListOrNull()
             is OrderedList -> toMarkdownListOrNull()
             is ThematicBreak -> MarkdownBlock.ThematicBreak
@@ -214,28 +207,20 @@ public class MarkdownProcessor(
                 extensions.find { it.processorExtension.canProcess(this) }
                     ?.processorExtension?.processMarkdownBlock(this, this@MarkdownProcessor)
             }
-
+            // TODO: add support for LinkReferenceDefinition
             else -> null
         }
 
     private fun BlockQuote.toMarkdownBlockQuote(): MarkdownBlock.BlockQuote =
         MarkdownBlock.BlockQuote(processChildren(this))
 
-    private fun Heading.toMarkdownHeadingOrNull(): MarkdownBlock.Heading? =
-        when (level) {
-            1 -> H1(contentsAsInlineMarkdown())
-            2 -> H2(contentsAsInlineMarkdown())
-            3 -> H3(contentsAsInlineMarkdown())
-            4 -> H4(contentsAsInlineMarkdown())
-            5 -> H5(contentsAsInlineMarkdown())
-            6 -> H6(contentsAsInlineMarkdown())
-            else -> null
-        }
+    private fun Heading.toMarkdownHeadingOrNull(): MarkdownBlock.Heading =
+        MarkdownBlock.Heading(contentsAsInlineMarkdown(), level)
 
-    private fun Paragraph.toMarkdownParagraphOrNull(): MarkdownBlock.Paragraph? {
+    private fun Paragraph.toMarkdownParagraphOrNull(): MarkdownBlock.Paragraph {
         val inlineMarkdown = contentsAsInlineMarkdown()
 
-        if (inlineMarkdown.isBlank()) return null
+//        if (inlineMarkdown.isEmpty()) return null
         return MarkdownBlock.Paragraph(inlineMarkdown)
     }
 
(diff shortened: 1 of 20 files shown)


obask:
I thought that it makes scrolling errors more prominent, but it looks exactly the same as before... not sure if the bug is in CfD or the way we represent lists or remember some nodes... which is probably not the case, because these nodes don't change during rendering at all.

rock3r:
You need to run the `apiDump` task for this to go green :)

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -107,28 +100,20 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
                 extensions.find { it.processorExtension.canProcess(this) }
                     ?.processorExtension?.processMarkdownBlock(this, this@MarkdownProcessor)
             }
-
+            // TODO: add support for LinkReferenceDefinition
```
rock3r, line 103: Note that link reference definitions were purposefully ignored and inlined into their respective links, because they break the encapsulation of block-level models. I still think we should inline references and ignore LinkReferenceDefinitions, since they have no UI value nor representation.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -107,28 +100,20 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
...
-        }
+    private fun Heading.toMarkdownHeadingOrNull(): MarkdownBlock.Heading =
+        MarkdownBlock.Heading(contentsAsInlineMarkdown(), level)
 
-    private fun Paragraph.toMarkdownParagraphOrNull(): MarkdownBlock.Paragraph? {
+    private fun Paragraph.toMarkdownParagraphOrNull(): MarkdownBlock.Paragraph {
         val inlineMarkdown = contentsAsInlineMarkdown()
 
-        if (inlineMarkdown.isBlank()) return null
+//        if (inlineMarkdown.isEmpty()) return null
```
rock3r, line 116: This was done on purpose, too — there is no point in empty paragraphs for what the UI is concerned.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -141,17 +126,11 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
...
-        return MarkdownBlock.Image(destination.trim(), title.trim())
-    }
-
-    private fun BulletList.toMarkdownListOrNull(): UnorderedList? {
+    private fun BulletList.toMarkdownListOrNull(): MarkdownBlock.ListBlock.BulletList? {
         val children = processListItems()
         if (children.isEmpty()) return null
 
-        return UnorderedList(children, isTight, bulletMarker)
+        return org.jetbrains.jewel.markdown.MarkdownBlock.ListBlock.BulletList(children, isTight, bulletMarker)
```
rock3r, line 133: This FQN seems left here by mistake

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
```diff
@@ -30,78 +17,73 @@ public open class DefaultInlineMarkdownRenderer(rendererExtensions: List<Markdow
...
-                is Block -> {
-                    error("Only inline Markdown can be rendered to an AnnotatedString. Found: $child")
+                is InlineMarkdown.Image -> {
+                    appendInlineContent(
+                        INLINE_IMAGE,
+                        child.value.destination + "\n" + plainTextRenderer.render(child.value),
+                    )
                 }
+
+                is InlineMarkdown.CustomNode -> error("InlineMarkdown.CustomNode render is not implemented")
```
rock3r, line 79: I guess this is the entry point for inline renderer extensions when we build those...

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -99,20 +92,22 @@ public open class DefaultMarkdownBlockRenderer(
...
-            is H5 -> render(block, rootStyling.heading.h5)
-            is H6 -> render(block, rootStyling.heading.h6)
+            is Heading -> when (block.level) {
+                1 -> render(block, rootStyling.heading.h1)
+                2 -> render(block, rootStyling.heading.h2)
+                3 -> render(block, rootStyling.heading.h3)
+                4 -> render(block, rootStyling.heading.h4)
+                5 -> render(block, rootStyling.heading.h5)
+                6 -> render(block, rootStyling.heading.h6)
+                else -> error("$block")
```
rock3r, line 102: Would be nice to have a more meaningful error description here
```suggestion
                else -> error("Heading level ${block.level} not supported:\n$block")
```

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -127,83 +122,18 @@ public open class DefaultMarkdownBlockRenderer(
...
-            is H5 -> render(block, styling)
-            is H6 -> render(block, styling)
+        when (block.level) {
+            1 -> render(block, styling.h1)
+            2 -> render(block, styling.h2)
+            3 -> render(block, styling.h3)
+            4 -> render(block, styling.h4)
+            5 -> render(block, styling.h5)
+            6 -> render(block, styling.h6)
         }
```
rock3r, line 132: Missing the else case:
```suggestion
            6 -> render(block, styling.h6)
            else -> error("Heading level ${block.level} not supported:\n$block")
        }
```

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingExtraTest.kt
```diff
@@ -7,6 +7,18 @@ class MarkdownProcessorDocumentParsingExtraTest {
 
     private val processor = MarkdownProcessor()
 
+    @org.junit.Ignore
```
rock3r, line 10: Add import for `Ignore`

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingTest.kt
```diff
@@ -466,6 +454,7 @@ class MarkdownProcessorDocumentParsingTest {
         parsed.assertEquals(paragraph("  & © Æ Ď ¾ ℋ ⅆ ∲ ≧̸"))
     }
 
+    @Ignore("leading # is not supported yet")
```
rock3r, line 457: Is this tracked somewhere?

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingTest.kt
```diff
@@ -8368,7 +8359,7 @@ class MarkdownProcessorDocumentParsingTest {
         parsed.assertEquals(paragraph("\\* a \\*"))
     }
 
-    @org.junit.Ignore("need to update escapeInlineMarkdownChars implementation")
+    @Ignore("need to update escapeInlineMarkdownChars implementation")
```
rock3r, line 8362: Ditto, is this tracked?

rock3r (changes requested):
Forgot to mark as "request changes" :)

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
obask: yes, but I don't want to make this PR too big

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingTest.kt
obask: I actually rewrote how we do diffs in tests, so that's not needed

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingExtraTest.kt
obask: deleted. Sometimes I use it when I only need to ignore one test, because it makes it easier to resolve merges in the future

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingTest.kt
obask: deleted escapeInlineMarkdownChars implementation

rock3r (approved):
LGTM but has conflicts to solve first

PR #332, opened 2024-03-27
Commonmark with value classes

obask:
This PR includes both changes to block nodes as well as inline nodes.

For now the Processor is a noop, but it could be helpful later if we decide to offload some operations from the rendering pipeline. It would be easier to decide once we implement quick typing improvements.


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
```diff
@@ -30,78 +17,73 @@ public open class DefaultInlineMarkdownRenderer(rendererExtensions: List<Markdow
...
-            appendInlineMarkdownFrom(node, styling)
+            appendInlineMarkdownFrom(inlineMarkdown.iterator(), styling)
         }
 
     @OptIn(ExperimentalTextApi::class)
-    private fun Builder.appendInlineMarkdownFrom(node: Node, styling: InlinesStyling) {
-        var child = node.firstChild
-
-        while (child != null) {
+    private fun Builder.appendInlineMarkdownFrom(inlineMarkdown: Iterator<InlineMarkdown>, styling: InlinesStyling) {
```
rock3r, line 34: Passing an `Iterator` around is kinda weird, because you can't know its state in advance. I'd prefer if it was receiving an `Iterable` instead so it can own its iterator

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -127,83 +127,18 @@ public open class DefaultMarkdownBlockRenderer(
...
-            is H4 -> render(block, styling)
-            is H5 -> render(block, styling)
-            is H6 -> render(block, styling)
+        when (block.level) {
+            1 -> render(block, styling.h1)
+            2 -> render(block, styling.h2)
+            3 -> render(block, styling.h3)
+            4 -> render(block, styling.h4)
+            5 -> render(block, styling.h5)
+            6 -> render(block, styling.h6)
```
rock3r, line 136: Missing the else case

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -373,9 +305,10 @@ public open class DefaultMarkdownBlockRenderer(
                 .then(if (styling.fillWidth) Modifier.fillMaxWidth() else Modifier),
         ) {
             Column(Modifier.padding(styling.padding)) {
-                if (block.mimeType != null && styling.infoPosition.verticalAlignment == Alignment.Top) {
+                val mimeType = remember { MimeType.Known.fromMarkdownLanguageName(block.value.info) }
```
rock3r, line 308: Could this be a property or function on our Fenced Block model? Value classes can have properties and functions IIRC. Ideally, I'd like to have less arcane ways to access useful information in cases like this. Our value classes can decorate and enrich the CommonMark models to make it easier to build extensions

markdown/extension/gfm-alerts/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/github/alerts/GitHubAlertProcessorExtension.kt
```diff
@@ -40,19 +36,20 @@ public object GitHubAlertProcessorExtension : MarkdownProcessorExtension {
...
+//            if (children.isEmpty()) return null
+//
+//            return when (block) {
+//                is Caution -> Alert.Caution(children)
+//                is Important -> Alert.Important(children)
+//                is Note -> Alert.Note(children)
+//                is Tip -> Alert.Tip(children)
+//                is Warning -> Alert.Warning(children)
+//                else -> error("Unsupported custom block of type ${block.javaClass.name}")
+//            }
```
rock3r, line 52: Should we remove commented code?

markdown/extension/gfm-alerts/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/github/alerts/GitHubAlertProcessorExtension.kt
```diff
@@ -40,19 +36,20 @@ public object GitHubAlertProcessorExtension : MarkdownProcessorExtension {
...
-            return when (block) {
-                is Caution -> Alert.Caution(children)
-                is Important -> Alert.Important(children)
-                is Note -> Alert.Note(children)
-                is Tip -> Alert.Tip(children)
-                is Warning -> Alert.Warning(children)
-                else -> error("Unsupported custom block of type ${block.javaClass.name}")
-            }
+        override fun processMarkdownBlock(block: CustomBlock, processor: MarkdownProcessor): MarkdownBlock.CustomBlock? {
+            error("should use CustomNode parser instead")
```
rock3r, line 40: I don't love APIs that have functions you must override but throw when you call them. Especially if the return type is nullable, wouldn't you just return null? Am I missing something?

markdown/extension/gfm-alerts/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/github/alerts/GitHubAlertBlockExtensionTest.kt
```diff
@@ -44,8 +39,8 @@ class GitHubAlertBlockExtensionTest {
 
         val firstChild = children.first()
         assertTrue(
-            "Parsed node should be a Note node, but was ${firstChild.javaClass.name}",
-            firstChild is Note,
+            "Parsed node should be a Note node, but was $firstChild",
```
rock3r, line 42: I think this message (and all similar following ones) should print firstChild.kind, not the whole firstChild

rock3r (changes requested):

obask:
addressed comments in https://github.com/JetBrains/jewel/commit/76b52827fbae2ebd6b42c8b31ba825c2853c958f but I can't create a chain of PRs, because I can't push to my branches

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
obask: Iterator worked until I made things nullable... so I ended up simply using a `List`

rock3r:
@obask you have write permissions, that shouldn't happen. What does the error say? (we can troubleshoot in chat)

PR #334, opened 2024-03-29
Add an autolink extension for Markdown

jrlogsdon:
Exposes the autolink extension offered by the commonmark java library, to markdown rendering in Compose. Since links are already supported for processing and rendering, we only need to hookup the parserExtension to get this to work.


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/MarkdownProcessorExtension.kt
```diff
@@ -14,20 +14,20 @@ public interface MarkdownProcessorExtension {
      * A CommonMark [ParserExtension] that will be used to parse the extended
      * syntax represented by this extension instance.
      */
-    public val parserExtension: ParserExtension
+    public val parserExtension: ParserExtension?
```
rock3r, line 17: Please update the KDocs to mention what the nullability means

rock3r (commented):
Thanks for your first PR! I think it's a-ok, however I'd update the KDocs, and maybe also use the new extension in the standalone demo for funsies

jrlogsdon:
Closing in favor of: https://github.com/JetBrains/jewel/pull/341

PR #335, opened 2024-03-29
Diff lines when deciding what blocks to update

obask:
Also add a parameter to not apply optimizations to documents which could be modified.

Added workarounds for LinkReferenceDefinition which shouldn't be a top level node.

### DIFF
--- a/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
+++ b/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.jewel.markdown.processing
 
+import org.commonmark.node.Block
 import org.commonmark.node.BlockQuote
 import org.commonmark.node.BulletList
 import org.commonmark.node.Code
@@ -23,6 +24,7 @@ import org.commonmark.node.SoftLineBreak
 import org.commonmark.node.StrongEmphasis
 import org.commonmark.node.Text
 import org.commonmark.node.ThematicBreak
+import org.commonmark.parser.IncludeSourceSpans
 import org.commonmark.parser.Parser
 import org.commonmark.renderer.text.TextContentRenderer
 import org.intellij.lang.annotations.Language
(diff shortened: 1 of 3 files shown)


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -41,19 +43,35 @@ import org.jetbrains.jewel.markdown.MimeType
...
+    private val optimizeEdits: Boolean = true,
+) {
 
     public constructor(vararg extensions: MarkdownProcessorExtension) : this(extensions.toList())
 
-    private val commonMarkParser =
-        Parser.builder().extensions(extensions.map { it.parserExtension }).build()
+    private val commonMarkParser = Parser.builder()
+        .extensions(extensions.map { it.parserExtension })
+        .also {
```
rock3r, line 60: Since this is a builder, you should technically use a `let` or `run` since they return the result of their lambdas

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -83,11 +101,87 @@ public class MarkdownProcessor(private val extensions: List<MarkdownProcessorExt
...
+        }
+        val (previousLines, previousBlocks, previousIndexes) = currentState
+        val newLines = rawMarkdown.lines()
+        val nLinesDelta = newLines.size - previousLines.size
+        // find a block prior to the first one changed in case some elements merge during the update
+        var firstBlock = 0
+        var firstLine = 0
+        var currFirstBlock = 0
+        var currFirstLine = 0
+        outerLoop@ for ((i, spans) in previousIndexes.withIndex()) {
```
rock3r, line 117: Been a long time since I've seen a label in Kotlin 🤣

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessorTest.kt
```diff
@@ -0,0 +1,55 @@
+package org.jetbrains.jewel.markdown.processing
+
+import org.jetbrains.jewel.markdown.BlockWithInlineMarkdown
+import org.jetbrains.jewel.markdown.MarkdownBlock
+import org.junit.Assert.assertArrayEquals
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class MarkdownProcessorTest {
```
rock3r, line 9: For such a big feature I would expect a ton of tests, can you please add more covering all possile cases?

E.g.:
 * First block edited
 * Last block edited
 * Block in the middle edited
 * Blocks merged
 * Blocks split
 * Blocks deleted
 * Blocks added

Etc

rock3r (changes requested):

markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessorTest.kt
obask: PTAL https://github.com/JetBrains/jewel/pull/336

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
obask: I see labels all the time because of nested lambdas and context choreography common in Kotlin. I'm sure it's a nightmare for Lint writers, but works fine in a compiler.

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
obask: I don't see much difference, but changed to `let`

PR #338, opened 2024-03-31
Clean up Commonmark API conventions

obask:
- use one node type for Heading
- rename 'content' to 'children' if it points to other blocks

Also clean up tests to not use parenthesis,
becasue it breaks navigation in IDEA

### DIFF
--- a/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
+++ b/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
@@ -2,9 +2,7 @@ package org.jetbrains.jewel.markdown
 
 public sealed interface MarkdownBlock {
 
-    public data class BlockQuote(val content: List<MarkdownBlock>) : MarkdownBlock
-
-    public interface CustomBlock : MarkdownBlock
+    public data class BlockQuote(val children: List<MarkdownBlock>) : MarkdownBlock
 
     public sealed interface CodeBlock : MarkdownBlock {
 
@@ -20,6 +18,8 @@ public sealed interface MarkdownBlock {
         ) : CodeBlock
     }
 
+    public interface CustomBlock : MarkdownBlock
+
     public data class Heading(
         override val inlineContent: List<InlineMarkdown>,
         val level: Int,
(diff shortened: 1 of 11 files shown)


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -222,15 +216,8 @@ public class MarkdownProcessor(
...
-            1 -> H1(contentsAsInlineMarkdown())
-            2 -> H2(contentsAsInlineMarkdown())
-            3 -> H3(contentsAsInlineMarkdown())
-            4 -> H4(contentsAsInlineMarkdown())
-            5 -> H5(contentsAsInlineMarkdown())
-            6 -> H6(contentsAsInlineMarkdown())
-            else -> null
-        }
+        MarkdownBlock.Heading(contentsAsInlineMarkdown(), level)
+            .takeIf { level in 1..6 }
```
rock3r, line 220: This is a pretty expensive check — we first convert the inline markdown, then create the Heading instance, then the 1..6 IntRange, and then we may be throwing it all away. A classic if would be better IMO:

```kotlin
private fun Heading.toMarkdownHeadingOrNull(): MarkdownBlock.Heading? {
  if (level < 1 || level > 6) return null
  return MarkdownBlock.Heading(contentsAsInlineMarkdown(), level)
}
```

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
```diff
@@ -52,14 +52,18 @@ public class MarkdownStyling(
         public val h6: H6,
     ) {
 
+        public interface HN : WithInlinesStyling, WithUnderline {
```
rock3r, line 55: Please make it sealed since it's not meant to be extended further

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
```diff
@@ -7,44 +7,31 @@ public sealed interface MarkdownBlock {
...
-            override val items: List<ListItem>,
+            override val children: List<ListItem>,
             override val isTight: Boolean,
             val startFrom: Int,
             val delimiter: Char,
         ) : ListBlock
 
-        public data class UnorderedList(
-            override val items: List<ListItem>,
+        public data class BulletList(
```
rock3r, line 26: I'm not a fan of this change, to be honest, and it was intentionally NOT called a `BulletList`. It's merely a question of taste from me, and I won't stop the PR because of it, since CommonMark also uses this poor naming 🙃

rock3r (changes requested):

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
obask: I don't know why Idea added squiggly lines suggested an auto fix for this(I don't really trust compiler optimizations) but I like a simple if statement better too

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
obask: Changed it back since it's a more common name

rock3r (approved):

PR #348, opened 2024-04-09
Fix TAB key event swallowing from SelectableLazyColumn key event handler

fscarponi:
Now 'TAB' event is evaluated correctly when SLC is focussed, requesting FocusManager to move focus to the next focusable element

### DIFF
--- a/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
+++ b/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
@@ -20,12 +20,19 @@ import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusDirection
 import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.focus.focusRequester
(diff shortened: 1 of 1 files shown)


foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -67,12 +74,18 @@ public fun SelectableLazyColumn(
...
     }
+    val focusManager = LocalFocusManager.current
     val focusRequester = remember { FocusRequester() }
     LazyColumn(
         modifier = modifier.onFocusChanged { isFocused = it.hasFocus }
             .focusRequester(focusRequester)
             .focusable(interactionSource = interactionSource)
             .onPreviewKeyEvent { event ->
+                // Handle Tab key press to move focus to next item
+                if (event.type == KeyEventType.KeyDown && event.key == Key.Tab && !event.isShiftPressed) {
```
rock3r, line 85: Can we handle shift-tab to move to previous, too?

rock3r (changes requested):

foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
fscarponi: yea, almost costless

rock3r (approved):

PR #353, opened 2024-04-15
Add theme name to ThemeDefinition and JewelTheme

### LINKED ISSUE #351: Define a theme key/name
Sometimes, when in the IDE you switch between two dark-but-not-identical themes (e.g., Dark and Darcula), parts of the Jewel components such as Markdown don't get properly invalidated, since they're remembered against isDark as a proxy for the theme.

We need a theme "name" that can be used as key to properly cause recomposition when moving between themes with the same isDark

rock3r:
Names are supposed to be unique per each theme. In standalone, they're hardcoded to "IntUI Light" and "IntUI Dark", and in the bridge they match the Swing LaF name.

This allows us to fix a bug in Markdown in the bridge where switching between two LaFs with the same isDark value would cause the Markdown text not to recompose appropriately.

Fixes #351

### DIFF
--- a/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/MarkdownViewer.kt
+++ b/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/MarkdownViewer.kt
@@ -21,16 +21,15 @@ import java.net.URI
 
 @Composable
 internal fun MarkdownPreview(@Language("Markdown") rawMarkdown: String, modifier: Modifier = Modifier) {
-    val isDark = JewelTheme.isDark
-
-    val markdownStyling = remember(isDark) { MarkdownStyling.create() }
+    val themeKey = JewelTheme.name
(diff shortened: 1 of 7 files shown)


samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/MarkdownViewer.kt
```diff
@@ -21,16 +21,15 @@ import java.net.URI
 
 @Composable
 internal fun MarkdownPreview(@Language("Markdown") rawMarkdown: String, modifier: Modifier = Modifier) {
-    val isDark = JewelTheme.isDark
-
-    val markdownStyling = remember(isDark) { MarkdownStyling.create() }
+    val themeKey = JewelTheme.name
```
devkanro, line 24: It seems not be used

rock3r: Used in the next line as remember key

devkanro (approved):

PR #356, opened 2024-04-15
Set default color of Divider composable

YasserDbeis:
The current color and fallback style create a transparent appearance for the Divider under Classic UI's Darkula theme. This change fixes such issue.

Manually verified that the color is correct for both new and Classic UI in all available themes.

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Divider.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Divider.kt
@@ -22,7 +22,7 @@ import org.jetbrains.jewel.ui.theme.dividerStyle
 public fun Divider(
     orientation: Orientation,
     modifier: Modifier = Modifier,
-    color: Color = Color.Unspecified,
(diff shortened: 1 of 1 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Divider.kt
```diff
@@ -22,7 +22,7 @@ import org.jetbrains.jewel.ui.theme.dividerStyle
 public fun Divider(
     orientation: Orientation,
     modifier: Modifier = Modifier,
-    color: Color = Color.Unspecified,
+    color: Color = JewelTheme.globalColors.borders.normal,
```
rock3r, line 25: This is not the right place for this — defaults should be set in the DividerStyle.color

rock3r (changes requested):

rock3r:
Since it's been a few days, I took the liberty to close this and implement the proper fix in #362

YasserDbeis:
Thank you!

PR #366, opened 2024-04-25
Add onTextClick to MarkdownBlockRenderer

rock3r:
This is a workaround to the fact that `ClickableText` swallows all clicks, since its `onClick` lambda has no way of saying if it did handle a click.

Hopefully this won't be needed anymore when we move to Compose 1.7, which brings in a proper `LinkAnnotation`.

> [!IMPORTANT]
> Please don't merge until #365 has been merged


devkanro (approved):

devkanro:
LGTM, but with some hardcoded numbers, what's the source for these magic numbers?

rock3r:
The numbers are coming from the pre-existing code — I had to duplicate a bunch of logic on our side to make it easier for users to use. The numbers were derived from the `JBFont` typography presets for headings, and the Int UI specs.

PR #367, opened 2024-04-25
Add Markdown styling to theme

rock3r:
We were lacking a good, easy way to render Markdown, requiring a lot of ceremony. Now it's much easier, you just need to provide the styling, renderer, and processor as composition locals (with the defaults coming from ProvideMarkdownStyling). Then use either Markdown(), or LazyMarkdown(), and your content is displayed. Better!

I have updated the samples to show how it works.

> [!IMPORTANT]
> Please don't merge until #365 and #366 have been merged


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -70,71 +70,92 @@ import org.jetbrains.jewel.ui.Orientation.Horizontal
...
             }
         }
     }
 
     @Composable
-    override fun render(block: Paragraph, styling: MarkdownStyling.Paragraph) {
+    override fun render(
+        block: Paragraph,
+        styling: MarkdownStyling.Paragraph,
+        onUrlClick: (String) -> Unit,
```
obask, line 121: [nit] I don't know if you anticipate adding more click handlers, but I would probably wrap them into a data class so you only need to pass one parameter around. On the other hand, using separate parameters is more explicit and makes it easier to understand what the code does.

obask (approved):

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: Not expecting to. As mentioned in the description I hope we won't need the second lambda when we get 1.7.0, as it's essentially just a workaround for current `ClickableText` limitations

PR #381, opened 2024-05-08
Add text colours to GlobalColors

rock3r:
It's practical to have the "base" text colours in GlobalColors:

 * Normal
 * Selected
 * Disabled
 * Info
 * Error

They apply to labels and most components, but not every component (e.g., default buttons have "inverted" text colors).

Note that the text colour is used as content colour, but not directly in the styling. Besides, when initialising the stylings, we can't really read these values as they haven't been init yet.

This is mostly a convenience for users that want to set a content colour (e.g., to visually show some text as disabled/error).

Note: GlobalColors.paneBackground was renamed to panelBackground for consistency with the Swing color key naming.

### DIFF
--- a/foundation/api/foundation.api
+++ b/foundation/api/foundation.api
@@ -26,12 +26,12 @@ public abstract interface annotation class org/jetbrains/jewel/foundation/Genera
 public final class org/jetbrains/jewel/foundation/GlobalColors {
 	public static final field $stable I
 	public static final field Companion Lorg/jetbrains/jewel/foundation/GlobalColors$Companion;
-	public synthetic fun <init> (Lorg/jetbrains/jewel/foundation/BorderColors;Lorg/jetbrains/jewel/foundation/OutlineColors;JJLkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Lorg/jetbrains/jewel/foundation/BorderColors;Lorg/jetbrains/jewel/foundation/OutlineColors;Lorg/jetbrains/jewel/foundation/TextColors;JLkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun equals (Ljava/lang/Object;)Z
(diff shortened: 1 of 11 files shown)


samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/markdown/MarkdownPreview.kt
```diff
@@ -20,13 +20,13 @@ import androidx.compose.ui.unit.dp
...
-import org.jetbrains.jewel.intui.markdown.dark
-import org.jetbrains.jewel.intui.markdown.light
-import org.jetbrains.jewel.intui.markdown.styling.dark
-import org.jetbrains.jewel.intui.markdown.styling.extension.github.alerts.dark
-import org.jetbrains.jewel.intui.markdown.styling.extension.github.alerts.light
-import org.jetbrains.jewel.intui.markdown.styling.light
+import org.jetbrains.jewel.intui.markdown.standalone.ProvideMarkdownStyling
+import org.jetbrains.jewel.intui.markdown.standalone.dark
+import org.jetbrains.jewel.intui.markdown.standalone.light
+import org.jetbrains.jewel.intui.markdown.standalone.styling.dark
```
devkanro, line 26: These changes seem to be same with #379 and can be avoided by rebasing

devkanro (approved):

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/markdown/MarkdownPreview.kt
rock3r: Hmmm that's odd, I had rebased before pushing yesterday, this should not be in the changes. Anyway, I'll rebase now.

PR #391, opened 2024-05-27
Switch to IJP 242 EAP

### LINKED ISSUE #389: Move main branch to 242
Main needs to move to IJP 242 EAP, and 241 should be moved to a release branch.

rock3r:
This in turn requires:
1. Updating Kotlin to 1.9.24 (matching the bundled version)
2. Updating the IJP Gradle plugin to 2.0 (beta)
3. Updating Gradle (not mandatory, but why not)
4. Updating Compose to 1.6.20-dev1646
5. Reworking the IDE sample build.gradle.kts in the new 2.0 syntax
6. Bumping the JBR version to 21

Closes #389 — the `releases/241` branch was created from `main` after #390 was merged.

Note: we disabled the CodeQL checks because no matter what, they tried building with the JDK 11 instead of 21, which is now required. So long, CodeQL.


hamen (approved):
💪✌️

PR #394, opened 2024-06-05
Don’t use `toInt(unit: kotlin.time.DurationUnit)` function

morozkin:
Hi!

I noticed that `HorizontalScrollbar` uses `toInt(unit:)` function which requires `kotlin.time` package. Such requirements leads to the runtime linkage error.

```
loader constraint violation: when resolving method 
'int kotlin.time.Duration.toInt-impl(long, kotlin.time.DurationUnit)' 
the class loader com.intellij.ide.plugins.cl.PluginClassLoader of the current class,
org/jetbrains/jewel/ui/component/ScrollbarsKt, 
and the class loader com.intellij.util.lang.PathClassLoader for the method's defining class,
kotlin/time/Duration, have different Class objects for the type 
kotlin/time/DurationUnit used in the signature 
(org.jetbrains.jewel.ui.component.ScrollbarsKt is in unnamed module of
loader com.intellij.ide.plugins.cl.PluginClassLoader, parent loader 'bootstrap'; 
kotlin.time.Duration is in unnamed module of loader com.intellij.util.lang.PathClassLoader)
```

Other scroll bar components don't rely on it and their usage doesn't produce such error.

Some side questions:
* Why do `shape` and `hoverDurationMillis` store `MutableState` instead of plain values?
* Why isn't there `key` for `remeber` calls that are used for `shape` and `hoverDurationMillis`?

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Scrollbars.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Scrollbars.kt
@@ -17,7 +17,6 @@ import androidx.compose.ui.unit.dp
 import org.jetbrains.jewel.foundation.theme.JewelTheme
 import org.jetbrains.jewel.ui.component.styling.ScrollbarStyle
 import org.jetbrains.jewel.ui.theme.scrollbarStyle
-import kotlin.time.DurationUnit
 import androidx.compose.foundation.ScrollbarStyle as ComposeScrollbarStyle
 
 @Composable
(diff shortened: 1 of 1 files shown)


rock3r:
We've never had this error. Would you be able to provide more details in an issue, including the conditions under which it occurs?

morozkin:
This error is only relevant for IntelliJ plugin development.

You can add `HorizontalScrollbar(rememberScrollbarAdapter(scrollState))` [here](https://github.com/JetBrains/jewel/blob/135f009034cc84cb1e317152ccdb6621cdac71b3/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt#L77) and run plugin after that.

rock3r:
> This problem happens with this repository, with idea-plugin sample plugin

It doesn't happen to any of the contributors of this repo, otherwise we'd have addressed the issue. However, I am working on updating the dependencies to align with 241 (and preparing for 242, separately), and I think that would be the proper fix for issues caused by Kotlin version misalignment. What do you think?

morozkin:
It sounds good.

However, I don't like that implementations of the same logic differ within the same file.

Also could you please answer these questions regarding scroll bars:
-  Why do shape and hoverDurationMillis store MutableState instead of plain values?
-  Why isn't there key for remeber calls that are used for shape and hoverDurationMillis?

rock3r:
> However, I don't like that implementations of the same logic differ within the same file.

Agree there — we are planning a more comprehensive review of all the code and hopefully it should get rid of these inconsistencies. Would be great if you could file an issue for it, so we don't forget.

> Why do shape and hoverDurationMillis store MutableState instead of plain values?
I suppose that so they don't get recreated on every recomposition, since they are _derived_ from the style.

> Why isn't there key for remeber calls that are used for shape and hoverDurationMillis?
Likely an oversight :) Please mention this in the aforementioned issue!

rock3r:
FYI 0.19.6 is out, bumping Kotlin to 1.9.24

morozkin:
I can confirm that there is no problem on version `0.19.6`

PR #420, opened 2024-07-01
Fix default button focus outline & chip colors

rock3r:
* The default button was missing spacing between its bg and the focus outline, and outlined buttons were slightly off. Now the button border is drawn on top of the focus outline, like Swing does. It is now consistent with Swing both in standalone and bridge:
  standalone-light
  standalone-dark

* The chips were drawing a double-outline when focused and selected. Now they only draw the selected outline if they are both focused and selected.
* The chips didn't clearly signal the pressed and selected states. The background now reflects these states better (standalone only). There are no specs so we're going with a bit more freedom in the standalone version, as the Swing impl is quite lacking in polish still

 Light | Dark
 --- | ---
 chips-light | chips-dark


hamen (approved):
LGTM. Thanks 👍🏻

PR #421, opened 2024-07-01
Upgrade to Compose 1.7.0-dev1703

hamen:
Tested on Windows 11 as stand-alone and bridge.


rock3r (approved):
LGTM, works on macOS too

PR #423, opened 2024-07-01
Fix soft line breaks treated as hard line breaks

### LINKED ISSUE #380: Soft new line in Blockquote shouldn't be interpreted as hard new line
image

As seen in the screenshot above, the soft newline should not be treated as a hard newline. As far as I remember, this didn't use to happen before your changes, and is probably a regression.

rock3r:
Fixes #380


hamen (approved):
ChefsKissFrenchChefGIF

PR #430, opened 2024-07-03
Fix 242 icon loading by providing new key-based API

rock3r:
In IJP 242 the icon loading has changed yet again. Instead of relying on icon mappings like it used to, now New UI icon paths are hardcoded into the `CachedImageIcon` themselves. This breaks Jewel's ability to load New UI icons, resulting in old UI icons being shown instead.

With this PR, we set to address this issue, and design an API that improves upon the existing ones. We have created the new `IconKey`, that encapsulates an icon resource path in a better typed and fully portable way allowing the same code to be used both in standalone and in bridge mode.

We have a new generation task in the `ui` module that traverses an `AllIcons`-style class and an icon mappings json file (like `PlatfomIconMappings.json`), and creates an icon keys file with the same structure. These keys are then used to look up icons, simplifying the code. This mechanism should work for 233 and 241 as well, since it bypasses the need for relying on the platform icon mapping, until we get better APIs, hopefully soon.

The `Icon` composable now accepts `PainterHint`s and uses them when rendering the icons, too.

What used to look like this:

```kotlin
Icon(
  "actions/close.svg",
  iconClass = AllIcons::class.java,
  modifier = Modifier.border(1.dp, Color.Magenta),
  contentDescription = "An icon",
)

Icon(
  "icons/github.svg",
  iconClass = JewelIcons::class.java,
  modifier = Modifier.border(1.dp, Color.Magenta),
  contentDescription = "An icon",
)
```

Now looks like this:

```kotlin
PlatformIcon(AllIconsKeys.Actions.Close, "An icon")

Icon(
  IdeSampleIconKeys.gitHub,
  iconClass = IdeSampleIconKeys::class.java,
  modifier = Modifier.border(1.dp, Color.Magenta),
  contentDescription = "An owned icon",
)
```

Applying hints to `Icon`s and `PlatformIcon`s is also much easier now. It was:

```kotlin
val iconProvider =
    rememberResourcePainterProvider("icons/taskGroup.svg", StandaloneSampleIcons::class.java)

val badged by iconProvider.getPainter(Badge(Color.Red, DotBadgeShape.Default))
val strokedAndBadged by iconProvider.getPainter(Badge(Color.Red, DotBadgeShape.Default), Stroke(Color.White))

Icon(badged, "taskGroup")
Icon(strokedAndBadged, "taskGroup")
```

And now is:

```kotlin
PlatformIcon(AllIconsKeys.Nodes.ConfigFolder, "taskGroup", hint = Badge(Color.Red))

PlatformIcon(
  AllIconsKeys.Nodes.ConfigFolder,
  "taskGroup",
  hints = arrayOf(Stroke(Color.White), Badge(Color.Red)),
)
```

The sample code has been partially updated to the new API, but more work is needed to spread the change throughout existing components and their styling. That will come in a follow-up PR.


devkanro (approved):

jakub-senohrabek-jb:
To reduce dependency on IntelliJ platform, maybe current composition context (or something better) could hold something like IconProvider, which can resolve icon by id? Then when used as IJ plugin, it could use IntelliJIconKey.fromPlatformIcon as the IconProvider and in case of desktop jewel, it can use any kind of resource based thing, or it can be even customizable. The Icon api would then accept this id, maybe it can be even same, but we can make the Icon class nullable, to indicate it should be automatically resolved.

```kotlin
interface IconProvider {
	fun getIconByKey(key: string): ResourcePainter? // or maybe just Painter?
}
```

rock3r:
That was an earlier iteration on the design, but it depends on having IJP classes on the classpath (e.g., `AllIcons`), or using stringly-typed APIs, which are anti-goals for this change. To make it work, we'd still need to generate the keys anyway, so at that point, might as well go all the way with our keys to the API. Custom themes should still be able to use json icon mappings to provide custom icons if they so wish.

buildSrc/build.gradle.kts
```diff
@@ -23,6 +23,8 @@ dependencies {
     implementation(libs.kotlinx.binaryCompatValidator.gradlePlugin)
     implementation(libs.kotlinx.serialization.json)
     implementation(libs.poko.gradlePlugin)
+    implementation(libs.intellijPlatform.util.ui)
+    implementation(libs.intellijPlatform.icons)
```
lamba92, line 27: This will f you up eventually in unforeseeable ways in the future. I recommend to create a configuration in the plugin and passed the resolved configuration to the task that generates the icons. Have a look at [here](https://chatgpt.com/share/e011f4b5-6e0f-4de0-bc85-26b505d6ed17).

lamba92 (changes requested):

buildSrc/build.gradle.kts
rock3r: Why should it? It's used in a task in the same way as the other dependencies! If it's a problem for these it'll be for the others too

ui/build.gradle.kts
```diff
@@ -19,3 +20,10 @@ dependencies {
         exclude(group = "org.jetbrains.compose.material")
     }
 }
+
+intelliJIconKeysGenerator {
+    register("AllIcons") {
```
jakub-senohrabek-jb, line 25: What about other Icon classes?

rock3r: We've limited ourselves to All icons for this PR, but others can be generated by simply registering extra ones here. I don't have a list of classes to generate, so if you can procure one, we'd be golden.

Note that for now we only aim to generate classes for icons in the base platform, not in other IDEs, since those would probably need to be put in separate modules. It's perfectly doable, as long as we have a way to depend on an artefact containing that class. It'll be even easier if get the extra info we need directly in DummyIcon. Just a matter if adding more modules. Separate PRs material for sure tho.

buildSrc/build.gradle.kts
lamba92: So you are adding the classes directly to the build classpath, leaking them into every build file. This behaviour can quickly go out of hand as project increase sizes. I would recommend to "isolate" the classpath of such dependencies by using a configuration and `URLClassLoader` so that such classes won't spill into the build. 

If this same mechanism is used for other tasks, I strongly suggest to isolate them as well.

rock3r: (we discussed this offline and @lamba92 offered to do this change)

jakub-senohrabek-jb (approved):

PR #439, opened 2024-07-08
Replace icon paths with keys

rock3r:
This PR replaces all direct usages of icon paths and painter providers in the components' styling with icon keys. There is no visual or behaviour difference (except now icon mapping works in IJ 242 and the dropdown component looks correct under New UI).


samples/standalone/build.gradle.kts
```diff
@@ -21,6 +21,18 @@ dependencies {
...
+    toolchain {
+        vendor = JvmVendorSpec.JETBRAINS
+    }
+}
+
+kotlin {
+    jvmToolchain {
+        vendor = JvmVendorSpec.JETBRAINS
+    }
+}
```
rock3r, line 34: This was previously done in the `jewel` plugin, but it's not anymore as it's not the recommended approach anymore for IDE-bound code. However, we still need it for the standalone sample, so we can ensure it is run on the JBR.

hamen (approved):
NickMillerGIF

PR #440, opened 2024-07-09
Fix snapshot publishing version

rock3r:
We were overwriting published versions by mistake when a snapshot build was created. Two-fold bug: using the wrong version in the repo publishing, and a wrong setting in Space packages allowing for overwriting packages.


fscarponi:
@lamba92 FYI this will fix artifact’s overrides

fscarponi (approved):

PR #444, opened 2024-07-10
Allow customizing the CommonMark Parser

rock3r:
This is required by one of our customers, and it allows them to customize the handling of standard tags (e.g., remove images).

It is an opt-in behaviour and it does not impact existing Markdown handling code or behaviour.


hamen (approved):
👍🏻

PR #448, opened 2024-07-15
Fix colour keys in bridge

rock3r:
We had some en-GB key name spelling (Grey) which wasn't found at runtime since the actual palette uses en-US spelling (Gray). This PR fixes it. No issues were present in the standalone version. These cosmetic issues were until now masked by us providing defaults that the code falls back on when it can't load the keys; we had a warning in the logs but I suppose nobody paid attention to them since it all worked anyway 😅


hamen (approved):
IgnoreWarningAustinEvansGIF

PR #468, opened 2024-07-19
Upgrade IJP Gradle plugin to 2.0.0-rc1

rock3r:
Fixes NoClassDefFoundExceptions when running the IDE sample with JBR 21.


rock3r:
Pushed a change since the IJP plugin version to use is `2.0.0-rc1` and not `2.0.0-RC1` (there was a minor naming issue on the IJP plugin side)

rock3r:
Removed Ivan as he is on vacation

jakub-senohrabek-jb (approved):

Issue #142, opened 2023-09-29
JBColor.namedColor() can't be used with a JBColor.marker()


rock3r:
In the `BridgeUtils`, we have to have this workaround to be able to map a missing theme colour to a null or a Compose `Color.Unspecified`:

```kotlin
private object FallbackMarker : JBColor(0x0000, 0x0000) {

    override fun toString() = "%%%%%%COLOR_NOT_FOUND_MARKER%%%%%%"

    override fun equals(other: Any?) = other === this

    override fun hashCode() = toString().hashCode()
}

@Suppress("UnstableApiUsage")
fun retrieveColorOrNull(key: String): Color? =
    JBColor.namedColor(key, FallbackMarker)
        .takeUnless { it.name == "NAMED_COLOR_FALLBACK_MARKER" || it == FallbackMarker }
        ?.toComposeColor()
```

If we use a `JBColor.marker()` as the default value for `namedColor()` instead of our own `FallbackMarker`, the equality check in the `takeUnless { }` fails and we end up trying to convert the marker to a Compose color, which will fail.

Theoretically it should work, but it doesn't. The ideal scenario would be to have a `JBColor.namedColorOrNull()`, but it's not there.

---

*Context:*

In Compose, we use `Color.Unspecified` when a color is not set, but Swing is assuming that, if no color is specified, it should return a fallback (which can be transparent, or something else, depending on the name of the key). That's where the default color, and `JBColor.marker()`, should come in — but we can't use marker because it doesn't work, and we don't have the actual default to use at the point where we're trying to read from the LaF.

rock3r:
@Walingar updates on this?

Walingar:
For now there is no better way to use this. So, let's keep current approach

rock3r:
@Walingar do you reckon filing an issue against the platform would be a good strategy?

Walingar:
@rock3r I think so, yes

rock3r:
Done — https://youtrack.jetbrains.com/issue/IDEA-345093/Add-a-way-to-detect-missing-named-colour-to-JBColor

rock3r:
@jakub-senohrabek-jb leaving it to you

PR #493, opened 2024-07-24
Use namedColorOrNull instead of marker for color retrieval (#142)

jakub-senohrabek-jb:
Related issue: https://github.com/JetBrains/jewel/issues/142


rock3r (approved):
Nice! Thanks Kuba :)

This is only possible on 242+, correct?

PR #498, opened 2024-07-26
Lower target JDK for 242

rock3r:
We had set 21 as that was the JDK that the IJP 242 runs on, but that was causing issues in some other places. Since the target JDK is a _minimum_ level, code targeting 17 also works fine in 21. Because of this, we have no reason not to lower it back to 17.


rock3r:
Smoke tested locally, shipping without review (yolo) as I need this urgently, and it's a small change.

Issue #483, opened 2024-07-23
Switch reflection-based system in standalone demo to simpler mechanism


rock3r:
The Standalone demo app uses reflection to determine what to show, but it's brittle (had issues on Windows) and makes finding usages difficult. We should replace it with a simpler solution, like a list of pages or something.

PR #457, opened 2024-07-18
Scrollbars followup, Textareas, Textfields, and more

hamen:
* TextAread and TextFields are now based on Compose 1.7 implementation
* Tabstrip scrollbar have a new style

PR #502, opened 2024-07-29
Add basic shortcuts to Standalone L1 navigation screens

### LINKED ISSUE #462: Add shortcuts to change sections in standalone demo
Shortcuts:
 * 0 -> welcome
 * 1 -> components
 * 2 -> markdown

Make sure shortcuts are shown in the dropdown items.

hamen:
As a developer quality of life improvement, we can navigate to the three L1 screens using the keyboard:

```
Alt + W -> Welcome
Alt + M -> Markdown
Alt + C -> Components
```

Shortcuts are different from the originally suggested ones because those will conflict with the system ones I'm using on Windows and would make the quality of my life the same, if not worse 😄 

The dropdown menu will not show the keybindings for now. We will iterate on this after https://github.com/JetBrains/jewel/issues/483.

This work was initially done in https://github.com/JetBrains/jewel/pull/457 and extracted to reduce the PR size.

### DIFF
--- a/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
+++ b/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
@@ -1,6 +1,12 @@
 package org.jetbrains.jewel.samples.standalone
 
 import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.input.key.Key
+import androidx.compose.ui.input.key.KeyEvent
+import androidx.compose.ui.input.key.KeyEventType.Companion.KeyDown
+import androidx.compose.ui.input.key.isAltPressed
+import androidx.compose.ui.input.key.key
+import androidx.compose.ui.input.key.type
 import androidx.compose.ui.res.ResourceLoader
 import androidx.compose.ui.res.loadSvgPainter
 import androidx.compose.ui.text.font.FontFamily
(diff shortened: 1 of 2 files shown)


samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -68,14 +76,41 @@ fun main() {
...
     }
 }
 
+/*
+    Alt + W -> Welcome
+    Alt + M -> Markdown
+    Alt + C -> Components
+ */
+private fun processKeyShortcuts(it: KeyEvent) =
+    when {
```
rock3r, line 95: Nit, but you could check the type and modifiers before the when, and straight up return false if Alt is not pressed or the event is not KeyDown

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -68,14 +76,41 @@ fun main() {
...
+        it.isAltPressed && it.key == Key.W && it.type == KeyDown -> {
+            currentView = views.first { viewInfo -> viewInfo.title == "Welcome" }
```
rock3r, line 97: Took me some time to figure out where `currentView` is coming from. You should not read/write global state directly; instead you should hoist an `onNavigateTo()` event.

I know the ViewModel in this case is global state, but I'd like one day to make it an actual ViewModel, now that we can actually do that :)

rock3r (approved):

Issue #504, opened 2024-07-29
Compose 1.7 breaks ComposePanel


francisconoriega:
I found a pretty odd bug in compose 1.7 that doesn't repro in 1.6.11:

When using a `Row` + `Modifier.fillMaxWidth` or a `Column` + `fillMaxHeight`, the bounds of the panel are broken whenever you change tabs or click on another panel's icon. Resizing the ComposePanel seems to fix things again.

Weirdly enough, for AS it only seems to happen if you also have a bottom panel open.

**This issue seems to be more in AS and/or Compose 1.7 itself, but I wonder if this means jewel should go back to 1.6.11 for stability**

Here's a minimalist example I manage to get that repros the issue:

``` 
internal class JewelDemoToolWindowFactory : ToolWindowFactory, DumbAware {
    override fun createToolWindowContent(project: Project, toolWindow: ToolWindow) {
        val composePanel = ComposePanel()
        composePanel.setContent {
            Column(modifier = Modifier.border(2.dp, Color.Magenta)) {
                Row(
                    modifier = Modifier.fillMaxWidth(), horizontalArrangement = Arrangement.End
                ) {
                    Box(Modifier.size(50.dp).background(Color.Green))
                }
                Column(
                    modifier = Modifier.fillMaxHeight(), verticalArrangement = Arrangement.Bottom
                ) { Box(Modifier.size(50.dp).background(Color.Green)) }
            }
        }
        toolWindow.contentManager.apply {
            val content = factory.createContent(composePanel, null, false)
            addContent(content)
        }
    }
}
```

Repro video:

https://github.com/user-attachments/assets/63e55b5b-07f3-46c8-ad3a-34b1c3c61543

francisconoriega:
it might be related to https://github.com/JetBrains/compose-multiplatform/issues/3206

francisconoriega:
I've opened https://youtrack.jetbrains.com/issue/CMP-5856/Desktop-ComposePanel-size-breaks-with-.fillMax-modifiers on the Jetbrains side, but I still think it would be worth considering downgrading to 1.6 in jewel for the meantime

rock3r:
Hi @francisconoriega, thanks for the thorough report! I will escalate with the Compose Multiplatform team. I can confirm I can also reproduce the issue, although it's intermittent and not happening 100% of the time.

Since we rely on some APIs introduced in 1.7, since 0.20, we can't downgrade to Compose 1.6.

francisconoriega:
[...]

PR #509, opened 2024-08-01
Compose 1.7 FillMaxSize+Jpanels Workaround

francisconoriega:
Workaround until the issue with Compose 1.7 + fillMax__ + IntelliJ Panels is fixed. 

It creates a box with a fixed required size at the root of `JewelComposePanel`, and uses the localcomposition to get the current panel (window) size.
 
 - https://github.com/JetBrains/jewel/issues/504
 - https://youtrack.jetbrains.com/issue/CMP-5856


rock3r (approved):
Thanks for this workaround! Hopefully this won't be needed for too long :)

PR #531, opened 2024-08-08
Expose PopupProperties in PopupMenu

francisconoriega:
PopupMenu internally uses Popup, but uses a hardcoded PopupPropterties object. This exposes it as a param and uses the old hard coded values as the default parameter value.


rock3r (approved):

rock3r:
@francisconoriega you need to run apiDump :)

francisconoriega:
> @francisconoriega you need to run apiDump :)

I thought it was an automatic CI step :P.

Would it make sense to run the klint, api-dump, and other tasks as as pre-commit hooks?

rock3r:
@francisconoriega it's not automatic because we want to ensure people are aware they're changing the APIs and do it knowingly. When we reach 1.0 we won't allow breaking changes anymore, unless it's a major change.

>Would it make sense to run the klint, api-dump, and other tasks as as pre-commit hooks?
I have a pre-push hook locally that runs ktlintCheck, we should eventually add it as a built-in. I looked into ways to do that "properly" in the past but never settled on one solution. Do you know any Gradle plugin to manage and install hooks?

I don't love the idea of running it as pre-commit as it's very slow. Plus, since we squash, we only care about the formatting when we push.

PR #532, opened 2024-08-09
Migrated ActionSystem to Compose Traversal API & UiDataProvider API

jakub-senohrabek-jb:
The api now looks like this:
```kotlin
val state = rememberTextFieldState("")
TextField(
    state = state,
    modifier = Modifier
        .width(200.dp)
        .provideData {
            set(ActionSystemTestAction.COMPONENT_DATA_KEY, "TextField")
            lazy(ActionSystemTestAction.COMPONENT_DATA_KEY) { 
                Math.random().toString()
            }
        },
    placeholder = { Text("Write something...") },
)
```
Let me know if the api can be improved somehow. ComponentDataProviderBridge is no longer public as it requires instance of JewelComposePanel and Swing level nesting. If there is reasonable use-case for this, we can make it public again.

### DIFF
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
@@ -11,19 +11,23 @@ import androidx.compose.ui.awt.ComposePanel
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalWindowInfo
 import androidx.compose.ui.unit.toSize
+import com.intellij.openapi.actionSystem.DataSink
+import com.intellij.openapi.actionSystem.UiDataProvider
 import org.jetbrains.jewel.bridge.actionSystem.ComponentDataProviderBridge
 import org.jetbrains.jewel.bridge.theme.SwingBridgeTheme
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.foundation.InternalJewelApi
+import java.awt.GridLayout
 import javax.swing.JComponent
+import javax.swing.JPanel
 
 @Suppress("ktlint:standard:function-naming", "FunctionName") // Swing to Compose bridge API
 public fun JewelComposePanel(content: @Composable () -> Unit): JComponent =
-    ComposePanel().apply {
+    createJewelComposePanel { jewelPanel ->
         setContent {
             SwingBridgeTheme {
-                CompositionLocalProvider(LocalComponent provides this@apply) {
-                    ComponentDataProviderBridge(this@apply, content = content)
+                CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+                    ComponentDataProviderBridge(jewelPanel, content = content)
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
```diff
@@ -32,18 +36,35 @@ public fun JewelComposePanel(content: @Composable () -> Unit): JComponent =
...
 @Suppress("ktlint:standard:function-naming", "FunctionName") // Swing to Compose bridge API
 public fun JewelToolWindowComposePanel(content: @Composable () -> Unit): JComponent =
-    ComposePanel().apply {
+    createJewelComposePanel { jewelPanel ->
         setContent {
             Compose17IJSizeBugWorkaround {
                 SwingBridgeTheme {
-                    CompositionLocalProvider(LocalComponent provides this@apply) {
-                        ComponentDataProviderBridge(this@apply, content = content)
+                    CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
```
jakub-senohrabek-jb, line 43: Should I set LocalComponent to the "Jewel wrapper" or leave it as ref to ComposePanel()?

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/ComponentDataProviderBridge.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    Box(modifier = Modifier.then(rootDataProviderModifier).then(modifier)) {
+        content()
+    }
+
+    DisposableEffect(component) {
+        component.targetProvider = rootDataProviderModifier
+
+        onDispose {
+            if (component.targetProvider == rootDataProviderModifier) {
```
jakub-senohrabek-jb, line 26: Should I use AtomicReference instead?

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
rock3r: I think you can leave it to the ComposePanel, unless there are reasons the Compose code may need access to the JewelComposePanel

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/ComponentDataProviderBridge.kt
rock3r: If you're worried about threading, it should not be a problem, since this should all happen in the composition (which is on the UI thread)

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ActionSystemTestAction.kt
```diff
@@ -0,0 +1,18 @@
+package org.jetbrains.jewel.samples.ideplugin
+
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DataKey
+import com.intellij.openapi.diagnostic.Logger
+
+class ActionSystemTestAction : AnAction() {
+    private val logger = Logger.getInstance(ActionSystemTestAction::class.java)
```
rock3r, line 9: Nit:
```suggestion
    private val logger = thisLogger()
```

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/DataProviderNode.kt
```diff
@@ -3,53 +3,26 @@ package org.jetbrains.jewel.bridge.actionSystem
...
- * Holder for parent node of current [DataProviderNode]. So, each
- * [DataProviderNode] provides itself and read parent node. It allows
- * building tree of [DataProviderNode] and traverse it later on.
- *
- * @see ModifierLocalModifierNode
- */
-private val LocalDataProviderNode = modifierLocalOf<DataProviderNode?> { null }
+public interface DataProviderContext {
+    public fun <TValue: Any> set(key: DataKey<TValue>, value: TValue?)
+    public fun <TValue: Any> lazy(key: DataKey<TValue>, initializer: () -> TValue?)
```
rock3r, line 11: I don't love these function names, they don't feel super intuitive, but they map 1:1 onto the underlying Action system APIs, so 🤷 I suppose that's what we have to work with. Plus, it could be more confusing to use different terminologies at this point...

rock3r (commented):
Kuba, can we move the interfaces/modifier definition to the UI package and have a dummy no-op implementation in the standalone package? Otherwise we're creating a bridge-only API.

(I am aware this was only in the bridge to begin with, but it was never really used/supported before, and we're here to change that)

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanel.kt
```diff
@@ -32,18 +36,35 @@ public fun JewelComposePanel(content: @Composable () -> Unit): JComponent =
...
+                        ComponentDataProviderBridge(jewelPanel, content = content)
                     }
                 }
             }
         }
     }
 
+private fun createJewelComposePanel(config: ComposePanel.(JewelComposePanel) -> Unit): JewelComposePanel {
+    val jewelPanel = JewelComposePanel()
+    jewelPanel.layout = GridLayout()
```
Walingar, line 53: Let's use `BorderLayout` and add `composePanel` to the `BorderLayout.Center`

Walingar (approved):

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/actionSystem/DataProviderNode.kt
jakub-senohrabek-jb: Do you have some better proposal? Indeed, at the moment they are 1:1 to the underlaying api.

rock3r: Nah, it's fine to map the IJ API names

rock3r (approved):
If you can confirm the APIs can now be used from standalone code too (they will be no-op, that's fine) this is good to go for me

PR #534, opened 2024-08-09
Add GH Pull Request template

hamen:
I would like to have this to keep a stricter discipline with my PRs, without the need of having everything in mind.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,29 @@
+## Description
+<!--- Describe your changes in detail -->
+
+## Related Issue
+<!--- This project only accepts pull requests related to open issues. -->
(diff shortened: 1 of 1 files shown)


rock3r (commented):
I'm dreading having to fill this in every time I open a PR. I understand why you want this for yourself, but is it really necessary to impose it on everyone's PRs, too?

hamen:
I don't want to impose anything. I'm not sure what's everybody else's workflow, and I thought these would be steps useful to everybody opening PR. It's a common thing in many projects.

rock3r:
It's just... _very_ verbose. I'd be ok with 2-3 items to check, but this is overwhelming for me. If we can slim it down a bit, I think we could ship it.

hamen:
No worries. I'll keep it as a local personal checklist 👍🏻

PR #546, opened 2024-08-16
Migrate TitleBarView to new Icon API

hamen:
This PR improves consistency and removes a few warnings.

image
image
image
image


devkanro (approved):

rock3r:
Do they still look right even without the size hints? This is not a lossless change, the way you implemented it

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/StandaloneSampleIcons.kt
```diff
@@ -8,9 +8,13 @@ object StandaloneSampleIcons {
     val jewelLogo = PathIconKey("icons/jewel-logo.svg", StandaloneSampleIcons::class.java)
     val markdown = PathIconKey("icons/markdown.svg", StandaloneSampleIcons::class.java)
     val themeDark = PathIconKey("icons/darkTheme.svg", StandaloneSampleIcons::class.java)
+    val themeDark20 = PathIconKey("icons/darkTheme@20x20.svg", StandaloneSampleIcons::class.java)
```
rock3r, line 11: No, this is not correct. You need to use the `Size` `PainterHint`.

rock3r (changes requested):

hamen:
Thank you for the feedback. I changed the approach. I used a different overload and added the same icons as before. I think they look the same now.

rock3r:
The new approach is incorrect, you should _never_ have to create additional keys. Use `PainterHints` instead, they're there for this purpose.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/StandaloneSampleIcons.kt
hamen: OK. I'll check the method.

hamen:
I added the `PainterHint` and removed the unnecessary icon declaration. I also removed the `20x20` SVG files.
Considering that I took me a moment to figure out how to use the `Size` `PainterHint`, would it be useful to point the user to this example in the `README`?

rock3r:
Why did you remove the svg files? They were there for a reason...

As for the `Size` `PainterHint` in the readme, isn't there any mention of `PainterHint`s in the readme already? If there is, I don't think we should specifically add `Size`, since it's a hint like all others. If there is no mention at all of PainterHint

rock3r:
Sorry for closing, I mis-clicked

hamen:
I restored the icons in 0f812a21. I thought we would use the new approach from now on.

hamen:
> As for the `Size` `PainterHint` in the readme, isn't there any mention of `PainterHint`s in the readme already?

We mention it in https://github.com/JetBrains/jewel?tab=readme-ov-file#icon-runtime-patching, but the size example popped out from a test. It's just me having little knowledge about this topic and thinking about other people not familiar with this stuff.

rock3r:
>We mention it in https://github.com/JetBrains/jewel?tab=readme-ov-file#icon-runtime-patching, but the size example popped out from a test. It's just me having little knowledge about this topic and thinking about other people not familiar with this stuff.

I think the vague mention we have is ok for now, because if we start documenting one `PainterHint` we need to document _all_ of them, and that's not a priority right now

rock3r:
This has conflicts that need fixing with main.

And, re:

>I restored the icons in https://github.com/JetBrains/jewel/commit/0f812a2181fe9cdc236eb351c054c3dd895b1805. I thought we would use the new approach from now on.

We _are_ using the new approach. The Size painter hint looks for an image with the right size in its file name, and if there isn't one, just stretches the "base" one.

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/TitleBarView.kt
```diff
@@ -92,30 +92,34 @@ fun DecoratedWindowScope.TitleBarView() {
                     when (MainViewModel.theme) {
                         IntUiThemes.Light ->
                             Icon(
-                                "icons/lightTheme@20x20.svg",
-                                "Themes",
-                                StandaloneSampleIcons::class.java,
+                                key = StandaloneSampleIcons.themeLight,
+                                contentDescription = "Light",
+                                iconClass = StandaloneSampleIcons::class.java,
```
rock3r, line 97: Here and below, you don't need to specify the iconClass anymore, the keys have one embedded that's used by default.

rock3r:
That's how the icon loading works in the IJP in general; it is _somewhat_ [documented](https://plugins.jetbrains.com/docs/intellij/icons.html#png-format-deprecated) — however, as usual, not very well.

rock3r (approved):

PR #552, opened 2024-08-19
Improve OverrideDarkMode kdocs

### LINKED ISSUE #550: OverrideDarkMode doesn't work as expected
OverrideDarkMode doesn't really sets the composition to use darkmode values, which what I would expect.

Example, given:

```
// running in Light UI
Box(modifier = Modifier.background(Color.Black)
    OverrideDarkMode(isDark = true) {
       Text("Hello").
    }
}
```
I'd expect the text/LocalContentColor to be set to  to DarkPalette, grey 12:

image

but instead I get LightPalette grey1:
image

which is unreadable against a dark background

rock3r:
Now it makes it clear that OverrideDarkMode does not change the theme, and provides an example on how to do that.

Closes #550


hamen (approved):
Clear now.

PR #555, opened 2024-08-22
Improve string-based Icon deprecation

rock3r:
Adding the PathIconKey import and param names to the ReplaceWith makes the auto-upgrade seamless. It would result in red code otherwise.

Upgrade Icon in SwingComparisonTabPanel to new style.


samples/ide-plugin/src/main/kotlin/icons/IdeSampleIconKeys.kt
```diff
@@ -1,7 +1,8 @@
 package icons
 
-import org.jetbrains.jewel.ui.icon.IntelliJIconKey
+import org.jetbrains.jewel.ui.icon.PathIconKey
 
 object IdeSampleIconKeys {
-    val gitHub = IntelliJIconKey("icons/github.svg", "icons/github.svg", IdeSampleIconKeys::class.java)
+    val gitHub = PathIconKey("icons/github.svg", IdeSampleIconKeys::class.java)
```
rock3r, line 6: We don't have NewUI mappings for these, so a simple PathIconKey is enough

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
```diff
@@ -42,7 +42,11 @@ import java.io.InputStream
 
 @Deprecated(
     "Use the IconKey-based API instead",
-    ReplaceWith("Icon(PathIconKey(resource, iconClass), contentDescription, colorFilter, modifier, hints)"),
+    ReplaceWith(
```
rock3r, line 45: Param names are necessary for the resulting code to compile, especially with the vararg versions.

samples/ide-plugin/src/main/kotlin/icons/IdeSampleIconKeys.kt
hamen: Nice. Yesterday, I replaced another icon on one of my branches and I created like the gitHub one you are improving. Gotta improve mine as well 🙏🏻

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
hamen: 💯 It saves a bit of time. I was manually adding the names and then running the ReplaceWith.

hamen (approved):

PR #565, opened 2024-08-28
Added coloring to Markdown code blocks

francisconoriega:
Currently, Jewel won't do syntax highlighting to Markdown code blocks, even though IJ does.

I added by using a third party library, https://github.com/SnipMeDev/Highlights.

The main changes are:

- Both Fenced and Indented styling now take in a SyntaxTheme (from Highlights library), which is just a color definition.
- BridgeMarkdownStyling creates it's theme by looking at DefaultLanguageHighlighterColors.
- IntUiMarkdownStyling has to hard coded themes, which match the colors in IntUI figma kit
- The Fenced/Indented renders parse the string and use an annotated string builder to color the different parts based on the tokenization of the highlights library and the passed in theme.

Additionally:
- Fixed a bug where IDE Sample would crash if the current theme didn't have Blue color defined.


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -342,7 +349,7 @@ public open class DefaultMarkdownBlockRenderer(
                 .then(if (styling.fillWidth) Modifier.fillMaxWidth() else Modifier),
         ) {
             Text(
-                text = block.content,
+                text = getSyntaxHighlightedCode(block.content, styling.coloringTheme),
```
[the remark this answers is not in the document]
rock3r: Oh also we need to make sure that `SyntaxTheme` is considered at least `@Stable`, if not ideally `@Immutable` — if it is not, we should wrap it/replace it

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -251,14 +251,22 @@ private fun IconsShowcase() {
         }
 
         Box(
-            Modifier.size(24.dp).background(JewelTheme.colorPalette.blue(4), shape = RoundedCornerShape(4.dp)),
+            Modifier.size(24.dp)
+                .background(
+                    JewelTheme.colorPalette.blueOrNull(4) ?: JewelTheme.globalColors.panelBackground,
```
rock3r, line 256: In which circumstances would this be a problem? When using the Old UI?

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -381,6 +389,9 @@ private fun MarkdownExample() {
                 |```kotlin
                 |fun hello() = "World"
                 |```
+                |    val indented = "This is an indented code block"
+                |
+                |
```
rock3r, line 394:
```suggestion
                |
                |    val indented = "This is an indented code block"
                |
```

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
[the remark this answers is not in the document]
rock3r: Right — ideally we should distinguish parsing and applying actual highlighting. Like the rest of the Markdown architecture, the former can/should be done off of the UI thread (e.g., in a ViewModel/Service), and leave the UI with only the job of applying the colours to the string. The latter part is not dissimilar to how we handle InlineMarkdown, since AnnotatedString is not mutable and needs knowledge of the style to be constructed.

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
francisconoriega: I was testing the change with other themes, and it crashed because this was null

rock3r: Would you be ok with contributing this fix separately for higher visibility in the release notes/history?

rock3r: Thanks, much appreciated 🙏

francisconoriega:
After talking to @rock3r, I seems clear that this is not quite what the Jewel project is aiming for in respect of Code highlighting, so I'm going to close the PR.

rock3r:
Thank you for understanding, and once again, for helping out the project 🙏

tmaxxdd:
I am the author of mentioned library and I wanted to put few words here.

 I wanted thank you francisconoriega for bringing my library here :) 
 And I thankful rock3r for suggestions that can make Highlights better.
 
 Some of the mentioned problems are addressed and maybe in future my code will be ready to be part something bigger as this project!

PR #568, opened 2024-08-29
Fixed Null Pointer in IDE sample on non INTUI themes

francisconoriega:
The IDE sample was crashing  the Component Showcase tab, when using some 3rd party themes do the color not being available. Changed to `colorOrNull` and use a backup color


rock3r (approved):
Thanks!

PR #570, opened 2024-08-30
New SplitLayout

hamen:
This PR introduces the new `SplitLayout`. The implementation takes inspiration from the existing one and from the Accompanist `TwoPane` one.

# Main points

* Colors match Swing 
* The divider is draggable
* The divider is `1.dp` wide, with an invisible wider area around it to facilitate click & dragging
* The divider position can be remembered in upper components or in a `ViewModel`
* The divider cannot be focused
* It allows nested layouts
* It allows minimum widths. When the available width is less than the desired panes minimum width, it shrinks the panes, keeping the expected ratio. This mirrors the Swing behavior.


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -177,24 +177,30 @@ private fun SplitLayoutImpl(
...
             Box(
-                Modifier.let { modifier ->
-                        if (strategy.isHorizontal()) {
-                            modifier.fillMaxHeight().width(draggableWidth)
-                        } else {
-                            modifier.fillMaxWidth().height(draggableWidth)
+                Modifier.let {
+                        when {
+                            strategy.isHorizontal() -> it.fillMaxHeight().width(draggableWidth)
+                            else -> it.fillMaxWidth().height(draggableWidth)
```
rock3r, line 183: This looked better as an if. Not a fan of using `when` with a simple boolean clause

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -369,19 +376,20 @@ private fun verticalTwoPaneStrategy(gapHeight: Dp = 0.dp): SplitLayoutStrategy =
             state: SplitLayoutState,
         ): SplitResult {
             val layoutCoordinates = state.layoutCoordinates ?: return SplitResult(Orientation.Horizontal, Rect.Zero)
-            val availableHeight = layoutCoordinates.size.height
+            val availableHeight = layoutCoordinates.size.height.toFloat().coerceAtLeast(1f)
```
rock3r, line 379: I don't think `toFloat()` is necessary

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -341,19 +347,20 @@ private fun horizontalTwoPaneStrategy(gapWidth: Dp = 0.dp): SplitLayoutStrategy
             state: SplitLayoutState,
         ): SplitResult {
             val layoutCoordinates = state.layoutCoordinates ?: return SplitResult(Orientation.Vertical, Rect.Zero)
-            val availableWidth = layoutCoordinates.size.width
+            val availableWidth = layoutCoordinates.size.width.toFloat().coerceAtLeast(1f)
```
rock3r, line 350: I don't think `toFloat()` is necessary

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
hamen: I could do without here, but then I would need to cast it to float in the other places I use it 😬

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+                        if (strategy.isHorizontal()) {
+                            modifier.fillMaxHeight().width(draggableWidth)
+                        } else {
+                            modifier.fillMaxWidth().height(draggableWidth)
+                        }
+                    }
                     .draggable(
+                        orientation = orientation,
+                        state =
+                            rememberDraggableState { delta ->
```
rock3r, line 180: Extract `dragState` variable please

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+
+                                    val newPosition =
+                                        (dragStartPosition + clampedPosition - dragStartPosition).coerceIn(
+                                            minFirstSize,
+                                            maxSize,
+                                        )
+                                    state.dividerPosition = newPosition / size
+                                }
+                            },
+                        onDragStarted = { offset ->
```
rock3r, line 205: `offset` is unused... Seems suspicious

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
                     .draggable(
+                        orientation = orientation,
+                        state =
+                            rememberDraggableState { delta ->
+                                state.layoutCoordinates?.let { coordinates ->
+                                    val size =
+                                        if (strategy.isHorizontal()) coordinates.size.width else coordinates.size.height
+                                    val minFirstSize = with(density) { minFirstPaneSize.toPx() }
+                                    val minSecondSize = with(density) { minSecondPaneSize.toPx() }
+                                    val maxSize = size - minSecondSize
```
rock3r, line 186: This is more `maxPositionPx` than `maxSize` — we're talking about the maximum position the split line can take, not the size of anything.

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+    dividerStyle: DividerStyle,
+    state: SplitLayoutState,
 ) {
     val density = LocalDensity.current
-    var dividerY by remember { mutableStateOf(with(density) { initialDividerPosition.roundToPx() }) }
+    var dragStartPosition by remember { mutableStateOf(0f) }
+    var currentDragPosition by remember { mutableStateOf(0f) }
+    var isDragging by remember { mutableStateOf(false) }
+    val resizeCursor = if (strategy.isHorizontal()) Cursor(Cursor.E_RESIZE_CURSOR) else Cursor(Cursor.N_RESIZE_CURSOR)
+    val defaultCursor = Cursor(Cursor.DEFAULT_CURSOR)
```
rock3r, line 146: Instead of re-creating these on each composition, I would store a `PointerIcon` directly as a top-level private property:

```kotlin
private val HorizontalResizePointerIcon = PointerIcon(Cursor(Cursor.E_RESIZE_CURSOR))
private val VerticalResizePointerIcon = PointerIcon(Cursor(Cursor.N_RESIZE_CURSOR))
```

For default, you have `PointerIcon.Default`. Easier, faster. Then line 152 becomes:

```kotlin
.pointerHoverIcon(if (isDragging) resizePointerIcon else PointerIcon.Default))
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+                .onGloballyPositioned { coordinates -> state.layoutCoordinates = coordinates }
+                .pointerHoverIcon(PointerIcon(if (isDragging) resizeCursor else defaultCursor)),
         content = {
+            Box(Modifier.layoutId("first")) { first() }
+            Box(Modifier.layoutId("second")) { second() }
+
             val dividerInteractionSource = remember { MutableInteractionSource() }
-            first(Modifier.layoutId("first"))
+            val dividerOrientation = if (strategy.isHorizontal()) Vertical else Horizontal
+            val fillMaxDirection = if (strategy.isHorizontal()) Modifier.fillMaxHeight() else Modifier.fillMaxWidth()
```
rock3r, line 159: I'd call this `fillModifier`

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+            val orientation = if (strategy.isHorizontal()) Orientation.Horizontal else Orientation.Vertical
 
             Divider(
-                orientation = Orientation.Horizontal,
-                modifier = Modifier.fillMaxHeight().layoutId("divider"),
-                color = dividerColor,
-                thickness = dividerThickness,
-                startIndent = dividerIndent,
+                orientation = dividerOrientation,
+                modifier = Modifier.then(fillMaxDirection).layoutId("divider").focusable(false),
```
rock3r, line 164: You don't need `Modifier.then()`, just start the chain with the first modifier:

```suggestion
                modifier = fillMaxDirection.layoutId("divider").focusable(false),
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+                        }
+                    }
                     .draggable(
+                        orientation = orientation,
+                        state =
+                            rememberDraggableState { delta ->
+                                state.layoutCoordinates?.let { coordinates ->
+                                    val size =
+                                        if (strategy.isHorizontal()) coordinates.size.width else coordinates.size.height
+                                    val minFirstSize = with(density) { minFirstPaneSize.toPx() }
```
rock3r, line 184: When you convert to pixels, I find it useful to name the variables accordingly, so it's easier to keep track of which ones are pixels and which ones are not when reading the code. I would call this `minPositionPx`, and inline the other one since it's only used in trivial math on the next line.

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+                                            minFirstSize,
+                                            maxSize,
+                                        )
+                                    state.dividerPosition = newPosition / size
+                                }
+                            },
+                        onDragStarted = { offset ->
+                            isDragging = true
+                            state.layoutCoordinates?.let { coordinates ->
+                                dragStartPosition =
```
rock3r, line 208: I may be wrong, but I feel like we don't read dragStartPosition anywhere other than the above redundant logic...

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
+                                currentDragPosition = dragStartPosition
+                            }
+                        },
+                        onDragStopped = { isDragging = false },
                         interactionSource = dividerInteractionSource,
-                        orientation = ComposeOrientation.Vertical,
-                        state = rememberDraggableState { delta -> dividerY += delta.toInt() },
                     )
-                    .pointerHoverIcon(PointerIcon(Cursor(Cursor.N_RESIZE_CURSOR)))
+                    .pointerHoverIcon(PointerIcon(resizeCursor))
```
rock3r, line 220: I don't think this is necessary anymore

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
-                width = incomingConstraints.maxWidth,
-                height = availableHeight - actualDividerY + dividerPlaceable.height,
+    ) { measurables, constraints ->
+        val firstMeasurable = measurables.find { it.layoutId == "first" } ?: error("No first component found.")
+        val secondMeasurable = measurables.find { it.layoutId == "second" } ?: error("No second component found.")
+        val dividerMeasurable = measurables.find { it.layoutId == "divider" } ?: error("No divider component found.")
+        val dividerHandleMeasurable =
+            measurables.find { it.layoutId == "divider-handle" } ?: error("No divider-handle component found.")
+
+        // The aesthetics of this block is meh but we can't use !! operator
```
rock3r, line 232: I don't get this comment — why not just using `if (state.layoutCoordinates != null) ... else ...`?

You can then extract a `doLayout(...)` function, and a `notReadyLayout()` one for the else.

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -15,184 +18,407 @@ import androidx.compose.runtime.mutableStateOf
...
-        val secondComponentConstraints =
-            Constraints.fixed(
-                width = incomingConstraints.maxWidth,
-                height = availableHeight - actualDividerY + dividerPlaceable.height,
+    ) { measurables, constraints ->
+        val firstMeasurable = measurables.find { it.layoutId == "first" } ?: error("No first component found.")
+        val secondMeasurable = measurables.find { it.layoutId == "second" } ?: error("No second component found.")
+        val dividerMeasurable = measurables.find { it.layoutId == "divider" } ?: error("No divider component found.")
+        val dividerHandleMeasurable =
+            measurables.find { it.layoutId == "divider-handle" } ?: error("No divider-handle component found.")
```
rock3r, line 230: Move these closer to their usages, no need to have them at this wider scope

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
rock3r: You're right, had missed some usages that do need it!

PR #572, opened 2024-08-30
Format with ktfmt

rock3r:
Reformat the entire project with ktfmt. Does not really need to be reviewed, it's all automatic changes.

The MergeSarif task stopped working for some reason — will be fixed in #573


hamen (approved):
LGTM 🤣🚀

rock3r:
The failure is simply the SARIF CI job not finding a SARIF file since I disabled the task... I'll disable that job for now

PR #577, opened 2024-09-04
Simplify Node.readInlineContent() implementation

rock3r:
Instead of building an iterator, then transforming it to a list, it now directly builds a list from the content.

Fixes #458


hamen (approved):
Cool

PR #587, opened 2024-09-07
Use a line-height of 1.3x for the default text styles

### LINKED ISSUE #585: defaultTextStyle lineheight does't match IDE
When showing blocks of text, IJ seems to be using a larger lineheight (seems like the default is 22.sp but it scales with font size).

This can be seen in multiple areas such as the kdoc tooltip, or the full documentation panel. The defaultTextStyle has no lineheight set at all, and is much more compact.

I'm not sure if this is the correct logic, but if you multiply the default font size of 13 by 1.69, you get 22, which is what it seems to be using at that size.

image
image

rock3r:
It looks like Swing uses a line height that is 1.3 times the font size, but that is not expressed anywhere I could find. I got the value by experimentation and pixel counting, like a proper DF review would 🙃 Using this value seems to work at both default and h[0-5] text sizes.

Having a non-default line height is a bit of a pain for users, since they need to remember updating the line height too when they customize the font size in a text style; however, we can't do anything about it, and in general, folks should stick to the predefined text styles.

To help, we now provide a `TextStyle.copyWithSize()` function that automatically sets the correct line height, making this slightly less painful.

Comparison between Swing (top) and Compose (bottom):

 Before | After
 --- | ---
 image | image

The difference is subtle, but it's there. I pixel counted and they match exactly with Swing (with a 1px tolerance on bigger text sizes) across all predefined text styles.

This closes #585


ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -148,9 +148,17 @@ public fun retrieveArcAsCornerSizeWithFallbacks(vararg keys: String): CornerSize
...
+    colorKey: String? = null,
+    lineHeight: TextUnit = TextUnit.Unspecified,
+    bold: Boolean = false,
+    fontStyle: FontStyle = FontStyle.Normal,
+    size: TextUnit = TextUnit.Unspecified,
+): TextStyle {
     val baseColor = colorKey?.let { retrieveColorOrUnspecified(colorKey) } ?: Color.Unspecified
-    return retrieveTextStyle(fontKey, color = baseColor)
+    val resolvedStyle = retrieveTextStyle(fontKey, color = baseColor, lineHeight, bold, fontStyle, size)
+    return resolvedStyle.copy(lineHeight = resolvedStyle.fontSize * 1.3)
```
c5inco, line 161: Should this be extracted as a constant or something to make it clearer it's a magic number?

c5inco: Oh I see you have it in `Typography` in the other part of this PR.

c5inco (commented):
Left one comment on the multiplier in the bridge

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
rock3r: Yeah I forgot to use the constant here — fixed in 9a03b6f, also with the issue we were ignoring the passed in line height if it was set.

hamen (approved):

Issue #591, opened 2024-09-10
Text Component has wrong order for color


francisconoriega:
Or at least it is not what Material uses.

When trying to figure out colors material uses:

    val textColor = color.takeOrElse { style.color.takeOrElse { LocalContentColor.current } }

where as Jewel uses

    val textColor = color.takeOrElse { LocalContentColor.current.takeOrElse { style.color } }

this means that in the next example, in Material it is possible to have the icon be blue and text be red, but in Jewel it is not possible, since `LocalContentColor` always wins so you get  a blue icon and a blue text.

```
  CompositionLocalProvider(
      LocalTextStyle provides LocalTextStyle.current.copy(color = Color.Red),
      LocalContentColor provides Color.Blue,
  ) {
      IconAndText()
  }

  @Composable
  fun IconAndText() {
      Row {
          Icon()
          Text(
              text = text,
          )
      }
  }

```

its a very small PR I can submit if you agree with this

rock3r:
I remember we changed this to fix a bug where text was not being displayed properly. At a quick look, this comes up: https://github.com/JetBrains/jewel/pull/180/files#r1361761781

Feel free to open a PR, but please text _extensively_ to make sure nothing breaks, both in Standalone and Bridge!

PR #592, opened 2024-09-10
Add syntax highlighting to markdown preview

AlexVanGogen:
I decided to go with the `Flow` in highlighter API so that implementations might be able to react to their internal alterations of a color scheme and provide new styled strings accordingly. Another approach that was in my mind is `highlight` returning a single `AnnotatedString` and a separate flow in the API acting as a trigger to re-`highlight` the code. It would make implementations of `highlight` simpler but at the cost of leaking into the foundation (and not every implementation would need that). If you know how it can be done better with compose, please tell, I'm still new to this and would love to learn good practices :)

For standalone, there are no working highlighters beside no-op one that just wraps a string without applying any styles. 

Indented code blocks aren't supported as well, mainly because they aren't supported in JCEF 🙂 but also because it requires  using some language guesser (and it seems a point for a futher discussion).

### DIFF
--- a/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
+++ b/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
@@ -8,9 +8,6 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.withStyle
 import com.intellij.lang.Language
 import com.intellij.lang.LanguageUtil
-import com.intellij.openapi.components.Service
-import com.intellij.openapi.components.service
-import com.intellij.openapi.editor.colors.EditorColorsListener
 import com.intellij.openapi.editor.colors.EditorColorsManager
 import com.intellij.openapi.editor.colors.EditorColorsScheme
 import com.intellij.openapi.editor.markup.TextAttributes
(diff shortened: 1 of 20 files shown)


markdown/ide-laf-bridge-styling/src/main/kotlin/org/jetbrains/jewel/intui/markdown/bridge/BridgeProvideMarkdownStyling.kt
```diff
@@ -21,12 +24,14 @@ public fun ProvideMarkdownStyling(
     markdownProcessor: MarkdownProcessor = remember { MarkdownProcessor() },
     markdownBlockRenderer: MarkdownBlockRenderer =
         remember(markdownStyling) { MarkdownBlockRenderer.create(markdownStyling) },
+    codeHighlighter: CodeHighlighter = remember { NoOpCodeHighlighter },
```
rock3r, line 27: This defaulting to no-op is a bit confusing. I understand this means we'd need to have the project available, but maybe we should create a `ProvideCodeHighlighting(project)` API that one can use to easily set things up, too

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -312,13 +315,13 @@ private fun RowScope.ColumnTwo() {
...
 @Composable
-private fun MarkdownExample() {
+private fun MarkdownExample(project: Project) {
     var enabled by remember { mutableStateOf(true) }
     CheckboxRow("Enabled", enabled, { enabled = it })
 
     val contentColor = if (enabled) JewelTheme.globalColors.text.normal else JewelTheme.globalColors.text.disabled
     CompositionLocalProvider(LocalContentColor provides contentColor) {
-        ProvideMarkdownStyling {
+        ProvideMarkdownStyling(codeHighlighter = remember(project) { project.service<LexerBasedCodeHighlighter>() }) {
```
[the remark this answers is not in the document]
rock3r: As for the project not changing here, I'm trying to think in which scenario we'd have the project change under us. As far as I'm aware, in a given ide window the project is never changing (swing code definitely assumes it), so we probably don't need to key the remember to the project.

ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
```diff
@@ -0,0 +1,119 @@
...
+        textAttributes: TextAttributes?,
+        block: AnnotatedString.Builder.() -> Unit,
+    ) {
+        if (textAttributes == null) {
+            return block()
+        }
+        withStyle(textAttributes.toSpanStyle(), block)
+    }
+
+    private fun TextAttributes.toSpanStyle() =
```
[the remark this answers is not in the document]
rock3r: For the record, you can try something like this approach: https://github.com/saket/extended-spans

However, this is not very good for perf. The Compose text team is aware of this need we have to have richer spans. We'll probably want to limit ourselves to what can be done already until a proper API is in place.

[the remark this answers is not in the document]
rock3r: Yeah that's my FR :)

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -378,6 +376,25 @@ public open class DefaultMarkdownBlockRenderer(
...
+            LocalCodeHighlighter.current.highlight(content, mimeType).collectAsState(AnnotatedString(content))
+        CodeText(annotatedCode, styling)
+    }
+
+    @Composable
+    private fun CodeText(annotatedCode: AnnotatedString, styling: MarkdownStyling.Code.Fenced) {
+        Text(
+            text = annotatedCode,
+            style = styling.editorTextStyle,
+            color = styling.editorTextStyle.color.takeOrElse { LocalContentColor.current },
```
francisconoriega, line 391:
```suggestion       
```

This line might be unnecessary, text will automatically get the color from the Color orElse TextStyle orElse LocalContent. Passing color would be more in case you needed to override the style color.

rock3r: This was changed by #591, so you may need to rebase on main to get that fix

markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -339,35 +343,29 @@ public open class DefaultMarkdownBlockRenderer(
 
     @Composable
     override fun render(block: FencedCodeBlock, styling: MarkdownStyling.Code.Fenced) {
```
francisconoriega, line 345: How come IndendtedCodeBlocks aren't getting the highlight as well? Is it intentional?

rock3r: Indented code blocks don't specify a language, so they don't get highlighted. I think that's a common pattern

rock3r (approved):
Great job! Let's ship this :)

PR #593, opened 2024-09-10
Fix IconKeys generator task

### LINKED ISSUE #590: AllIconsKeys is missing a lot of icons
For example, all icons after Actions.MenuCut are missing. Is it because the next one, Menu_cut, is deprecated?

rock3r:
We had a `return` instead of `return@forEach` that made the visitor terminate early as soon as it met a deprecated field, instead of just skipping the deprecated fields.

This fixes #590


buildSrc/src/main/kotlin/icon-keys-generator.gradle.kts
```diff
@@ -149,7 +149,7 @@ open class IconKeysGeneratorTask : DefaultTask() {
 
                 if (field.annotations.any { it.annotationClass == java.lang.Deprecated::class }) {
                     logger.lifecycle("Ignoring deprecated field: $fieldName")
-                    return
+                    return@forEach
```
rock3r, line 152: This is the actual fix :)

jrlogsdon (approved):

PR #595, opened 2024-09-16
Update Compose to 1.7.0-beta02

rock3r:
This version of Compose deprecates the resource loading APIs we've been using, and brutally removes ResourceLoader, so we need to migrate to the new (experimental) CMP Resources API.

Things seem to work fine after upgrading and migrating. Would appreciate further testing on platforms other than macOS before merging.


hamen:
On it 👍

hamen:
Tested it on Windows, for Standalone and IDE: progress bar and icons seem to work OK.

hamen (approved):

PR #596, opened 2024-09-17
Fixed Text Component order for color

francisconoriega:
Addressing https://github.com/JetBrains/jewel/issues/591

Reverts Text component logic for getting the color to one that matches Material, which is one that respects the caller's  passed in parameters first (color, then the color in the textStyle), and then falls back to localContentColor.

The code originally used what I propose here, but was later swapped to solve a bug with Tooltip + LightTheme: https://github.com/JetBrains/jewel/pull/180/files#r1361761781, but I believe this is the correct fix.

I tested that tooltip still works on ide + standalone, on light and dark


rock3r (approved):
Thank you for testing! I suspect we had a bug that was masked by this change, but yeah it makes sense for the order to be this, here.

PR #598, opened 2024-09-17
Remove underline on unhovered Links

francisconoriega:
IJ's default markdown does *not* show underline on links when they are not hovered, it only shows them on hovered.  This matches that. 

Since we currently don't support the more advance rounded border span for Focused state, I also included the underline for Focus.

https://github.com/user-attachments/assets/66c188b8-4c7c-4f1b-afbf-55729c411e06

https://github.com/user-attachments/assets/d974f118-5897-4390-b49f-06836d183294

cc: @rock3r


rock3r (approved):
LGTM, thanks!

PR #602, opened 2024-09-20
Changed grey to gray in readPaletteColor

francisconoriega:
It seems that from 242 and onward, the system color name for the gray palette changed from `grey` to `gray`. We had previously changed the name of the variables but not the name of the value.

`Grey` works fine on 233 and 241 though, so I think this needs a new release to be setup so that the change only affects 242+

This fixes https://github.com/JetBrains/jewel/issues/603


rock3r (approved):
Huh, I am silly. Forgot the one most important place :D

I'll take care not to backport this for the next release. Thanks :)

PR #615, opened 2024-09-30
Fix z-order for text field/area placeholder

### LINKED ISSUE #610: InputField — placeholder should be drawn behind cursor
In the inputFields' layout code (in both the `TextField` and `TextArea` decorators) we should make sure we draw the placeholder _behind_ the BTF. This way the cursor will display in front of the placeholder text instead of behind it.

image

This should be as simple as ensuring the BTF has a higher z-order than the placeholder (either by setting an explicit zIndex, or implicitly by changing the order the components are added to the layout)

rock3r:
This changes the order in which the placeholder and actual text field are laid out in the TextField and TextArea code. Since the two have the same zIndex (the default value, 0f) their placement order is reflected in the drawing order: the last placed is drawn last, on top of the rest.

This fixes a minor cosmetic issue, where the placeholder would be drawn on top of the cursor when the TextArea/TextField is empty.

Fixes #610

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextArea.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextArea.kt
@@ -339,11 +339,11 @@ private fun TextAreaDecorationBox(
         val height = calculateHeight(textAreaPlaceable, placeholderPlaceable, incomingConstraints)
 
         layout(width, height) {
+            // Placed similar to the input text below
+            placeholderPlaceable?.placeRelative(0, 0)
+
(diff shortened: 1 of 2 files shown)


hamen (approved):

kateliu20:
Thank you!!

rock3r:
You're welcome 🤗 it's available in 0.25

kateliu20:
Is jewel 0.25 compatible with  IJ 241? Just want to make sure.

rock3r:
Yes. Just make sure you use the -241 artefacts. 

The supported IJP versions table in the readme is up to date

PR #626, opened 2024-10-10
Add Pointer Release Handler to PointerEventActions

abarrafo:
Supporting more granular control for consumers by exposing a `handlePointerEventRelease`

Includes example demonstrating Customized `DefaultTreeViewPointerEventAction` modifying long-press vs short-press, and, right click behavior.  Model use case: When you want to use context menu actions with many selectedElement(s). Or, when you want to support click, hold and drag-n-drop actions with selectedElement(s) without reseting the selection state.

### DIFF
--- a/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/ChipsAndTree.kt
+++ b/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/ChipsAndTree.kt
@@ -195,4 +195,16 @@ fun TreeSample(modifier: Modifier = Modifier) {
             Box(Modifier.fillMaxWidth()) { Text(element.data, Modifier.padding(2.dp)) }
         }
     }
+
+    Text("Tree with custom click handler")
+    Box(modifier.border(1.dp, borderColor, RoundedCornerShape(2.dp))) {
+        CustomLazyTree(
+            tree = tree,
+            modifier = Modifier.size(200.dp, 200.dp),
+            onElementClick = {},
+            onElementDoubleClick = {},
+        ) { element ->
+            Box(Modifier.fillMaxWidth()) { Text(element.data, Modifier.padding(2.dp)) }
+        }
+    }
 }
--- a/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
+++ b/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
@@ -1,0 +1,127 @@
+package org.jetbrains.jewel.samples.standalone.view.component
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.takeOrElse
+import androidx.compose.ui.input.pointer.PointerEvent
+import androidx.compose.ui.input.pointer.isPrimaryPressed
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.lazy.SelectableColumnKeybindings
+import org.jetbrains.jewel.foundation.lazy.SelectableLazyItemScope
+import org.jetbrains.jewel.foundation.lazy.SelectableLazyListKey
+import org.jetbrains.jewel.foundation.lazy.SelectableLazyListState
+import org.jetbrains.jewel.foundation.lazy.SelectionMode
+import org.jetbrains.jewel.foundation.lazy.tree.BasicLazyTree
+import org.jetbrains.jewel.foundation.lazy.tree.DefaultTreeViewKeyActions
+import org.jetbrains.jewel.foundation.lazy.tree.DefaultTreeViewPointerEventAction
+import org.jetbrains.jewel.foundation.lazy.tree.KeyActions
+import org.jetbrains.jewel.foundation.lazy.tree.PointerEventActions
+import org.jetbrains.jewel.foundation.lazy.tree.Tree
+import org.jetbrains.jewel.foundation.lazy.tree.TreeElementState
+import org.jetbrains.jewel.foundation.lazy.tree.TreeState
+import org.jetbrains.jewel.foundation.lazy.tree.rememberTreeState
+import org.jetbrains.jewel.foundation.theme.JewelTheme
+import org.jetbrains.jewel.foundation.theme.LocalContentColor
+import org.jetbrains.jewel.ui.component.Icon
+import org.jetbrains.jewel.ui.component.styling.LazyTreeStyle
+import org.jetbrains.jewel.ui.theme.treeStyle
+
+@ExperimentalJewelApi
+@Composable
+fun <T> CustomLazyTree(
+    tree: Tree<T>,
+    modifier: Modifier = Modifier,
+    onElementClick: (Tree.Element<T>) -> Unit = {},
+    treeState: TreeState = rememberTreeState(),
+    onElementDoubleClick: (Tree.Element<T>) -> Unit = {},
+    onSelectionChange: (List<Tree.Element<T>>) -> Unit = {},
+    keyActions: KeyActions = DefaultTreeViewKeyActions(treeState),
+    style: LazyTreeStyle = JewelTheme.treeStyle,
+    nodeContent: @Composable (SelectableLazyItemScope.(Tree.Element<T>) -> Unit),
+) {
+    val colors = style.colors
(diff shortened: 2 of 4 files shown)


samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
```diff
@@ -0,0 +1,127 @@
...
+                .contentFor(TreeElementState.of(focused = isActive, selected = isSelected, expanded = false))
+                .value
+                .takeOrElse { LocalContentColor.current }
+
+        CompositionLocalProvider(LocalContentColor provides resolvedContentColor) { nodeContent(it) }
+    }
+}
+
+// Example detecting long press vs short press to handle pointer events separately. Disable
+// selection here for both primary button presses, and long press (drag) use-cases
```
abarrafo, line 83:
```suggestion
// selection behavior here for both right button presses, and long press (drag) use-cases
```

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/ChipsAndTree.kt
```diff
@@ -195,4 +195,16 @@ fun TreeSample(modifier: Modifier = Modifier) {
...
         }
     }
+
+    Text("Tree with custom click handler")
+    Box(modifier.border(1.dp, borderColor, RoundedCornerShape(2.dp))) {
+        CustomLazyTree(
+            tree = tree,
+            modifier = Modifier.size(200.dp, 200.dp),
+            onElementClick = {},
+            onElementDoubleClick = {},
```
rock3r, line 205: These don't really seem too useful since they don't demonstrate much. Maybe add listeners on the existing tree sample, to show a menu or something?

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
rock3r: You need to make the changes to your branch if you want to tweak the comment :)

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.jetbrains.jewel.foundation.lazy.tree.rememberTreeState
+import org.jetbrains.jewel.foundation.theme.JewelTheme
+import org.jetbrains.jewel.foundation.theme.LocalContentColor
+import org.jetbrains.jewel.ui.component.Icon
+import org.jetbrains.jewel.ui.component.styling.LazyTreeStyle
+import org.jetbrains.jewel.ui.theme.treeStyle
+
+@ExperimentalJewelApi
+@Composable
+fun <T> CustomLazyTree(
```
rock3r, line 33: I don't think this is super valuable as-is. I get what you're trying to do, but I have some concerns about the way it is implemented here.

This adds extremely little to the existing LazyColumn APIs, and if it were that important to have a `SimplePointerEventActions` class, then it should live alongside the `SelectableLazyColumn`, and potentially be used in a `SelectableLazyColumn` and `LazyTree` overload, not be in a sample.

rock3r (changes requested):

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
abarrafo: Yup,  I figured. The change I really want  in is, `public fun handlePointerEventRelease()` on the PointerEventActions and subsequent call on the BasicLazyTreeColumn.  Allowing for us to handle `List<Elements>` in the tree with more granularity depending on the pointer events. 

Agreed this really does not add much outside of a very simple sample (demonstrating right click, and long click selection behavior overrides in comparison to the default). Maybe we update the Lazy tree to simply take a `PointerEventActions`, that would simplify things a bit for us users downstream wanting to handle those events.   

I'm not sure a sample is even value add, to be honest.

rock3r: I'll leave it to @fscarponi to decide on the Tree API, I have nothing against surfacing the PointerEventActions there.

fscarponi:
Hi Folks,

Something seems off. I need to dive into this topic a bit more deeply.
I'm currently swamped with work, so I'll need a bit of time to look into this. I should be able to take a first look tomorrow and hopefully provide you with some feedback then. Be patient!

samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/CustomLazyTree.kt
fscarponi: see [here](https://github.com/JetBrains/jewel/pull/626#issuecomment-2412031527)

abarrafo:
Here is a quick demo of the current behavior, although @fscarponi  already gave me a path for handling right clicks, i just have not implemented that yet on my file explorer. 

Essentially, the issue is we lose selected items on click and drag actions... Which the above pointer action addition would give me a path to address. 


https://github.com/user-attachments/assets/740f14d9-8b89-4158-ad51-b84cb445097e

fscarponi:
> Here is a quick demo of the current behavior, although @fscarponi already gave me a path for handling right clicks, i just have not implemented that yet on my file explorer.
> 
> Essentially, the issue is we lose selected items on click and drag actions... Which the above pointer action addition would give me a path to address.
> 
>  Screencast.from.2024-10-15.13-05-37.mp4

i'm investigating right now, thanks, the use case video will help me!

fscarponi:
Hi @abarrafo , Tanks for the patience! what if we consider handling this logic where all pointerEvents are handled?
I've not finished yet to cover this use case, but this is the idea:
```

public class DragTreeViewPointerEventAction(
    val treeState: TreeState,
    val onMoveElement: (Set<*>) -> Unit,
) : DefaultSelectableLazyColumnEventAction() {
    override fun handlePointerEventPress(
        pointerEvent: PointerEvent,
        keybindings: SelectableColumnKeybindings,
        selectableLazyListState: SelectableLazyListState,
        selectionMode: SelectionMode,
        allKeys: List<SelectableLazyListKey>,
        key: Any,
    ) {
        with(keybindings) {
            when {
                pointerEvent.keyboardModifiers.isContiguousSelectionKeyPressed &&
                    pointerEvent.keyboardModifiers.isCtrlPressed -> {
                }

                pointerEvent.keyboardModifiers.isContiguousSelectionKeyPressed -> {
                    super.onExtendSelectionToKey(key, allKeys, selectableLazyListState, selectionMode)
                }

                pointerEvent.keyboardModifiers.isMultiSelectionKeyPressed -> {
                    selectableLazyListState.lastKeyEventUsedMouse = false
                    super.toggleKeySelection(key, allKeys, selectableLazyListState, selectionMode)
                }

                else -> {

                    //drag and drop have sense only if u act on an already selected element
                    if (key !in selectableLazyListState.selectedKeys) {

                        selectableLazyListState.selectedKeys = setOf(key)

                    } else {

                        when (val eventType = pointerEvent.type) {

                            PointerEventType.Unknown, PointerEventType.Enter , PointerEventType.Move -> {
                                //not needed
                            }

                            PointerEventType.Press -> {
                                val pointerButton=pointerEvent.button ?: return@with
                                when {
                                    pointerButton.isPrimary ->{
                                        //do nothing this can be a drag n drop
                                    }
                                    pointerButton.isSecondary ->{
                                        //handle secondary click logic
                                    }
                                    else->{
                                        //go deep if u needs
                                    }
                                }
                            }
                            PointerEventType.Release ->{ //change the selection on release
                                selectableLazyListState.selectedKeys = setOf(key) //change the selection
                            }
                            PointerEventType.Exit ->{
                                //is a mouse button pressed? are u leaving the area? this is a drag and drop
                                if(primaryButtonIsDown()) onMoveElement(treeState.selectedKeys)
                            }
                            

                        }
                    }
                }
            }
        }
    }
}
```

abarrafo:
That should work.

abarrafo:
I think we want extend from DefaultTreeViewPointerEventAction instead? But otherwise I believe this will work.

abarrafo:
Yes, confirmed, this works. :bow:  

We can close this.


https://github.com/user-attachments/assets/b186c8da-3292-479d-b6d7-d144b53c598d

fscarponi:
@abarrafo We will be glad if you take time to write a small example for standalone

PR #627, opened 2024-10-14
Add ListComboBox

hamen:
# Overview

This PR contains new components that are meant to replace the `Dropdown` composable.

## ListComboBox

This is the simplest way to create an editable dropdown menu that shows a list and allows adding a value that is not present in the list. The API is very minimal to keep things "nice and easy":

```
public fun ListComboBox(
    items: List<String>,
    listItemContent: @Composable (String, Boolean, Boolean) -> Unit,
)
```

### ComboBox

This is one of the building blocks and represents the editable version of the dropdown component, containing the empty popup and the editable text field.

### NotEditableComboBox

This is one of the building blocks and represents the editable version of the dropdown component, containing the empty popup and the text label.

### SimpleListItem

This is a pre-made composable that can help to create a list `ComboBox` without stressing over the styling.

## PopupContainer

This is the empty popup used by the `ComboBox` that can be filled with custom composables.


rock3r:
When testing, I found these issues:
* Dark theme selection colour is wrong (it's the light theme one), at least in standalone
  image
* Selection doesn't move with hover
  https://github.com/user-attachments/assets/1c9212d2-2482-42e4-bc9e-d6dbbf84af1d
* In bridge, the caret icon is the wrong size
  image
* Text input changes selection when opening the popup, but it shouldn't
  https://github.com/user-attachments/assets/7ff5dd82-617e-4389-b1e7-7c17458a8c62
  
This is just a quick smoke test. I will not review this PR until these are fixed, as it's bound to introduce more changes (and hopefully not more bugs)

c5inco:
One small nit on behavior is with arrow keys. When using the Up Arrow to navigate through the list to the top, once you get to the top, the subsequent key press moves the insertion cursor in the `TextField` to the beginning.
arrowing-up

It's likely behavior inherited from using `TextField`, but looking at the Swing implementation, it doesn't do this. So curious if we should be consistent here.

c5inco:
I also notice this PR contains a handful of changes to other components that might not be directly related to the creation of this new component?

For example:
- All additions in `ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/*`
- Styling additions for `SelectableLazyColumn`, `PopupContainer`

hamen:
> When using the Up Arrow to navigate through the list to the top, once you get to the top, the subsequent key press moves the insertion cursor in the TextField to the beginning.

Nice catch. Your guess is plausible. I'll work to make it more consistent with Swing.

> I also notice this PR contains a handful of changes to other components that might not be directly related to the creation of this new component?

You are correct again. This PR is the result of too many weeks of work, and contains iterations on too many things. For example, this PR is blocking the work on Split Button, that was started and then stopped because the `PopupContainer` is in this PR and hasn't been merged to main yet.
To avoid these problems from happening again, we changed the workflow, to increase task granularity, to have smaller PRs, and hopefully faster delivery.

rock3r (approved):
This PR is large enough as it is, but now that all big problems have been addressed, let's ship this and iterate on polish off main

PR #635, opened 2024-10-16
Fix missing ⬇️ event on editable ComboBox

hamen:
To keep the level of noise manageable, I suggest a commit-by-commit review.

### DIFF
--- a/ui-tests/src/test/kotlin/ListComboBoxUiTest.kt
+++ b/ui-tests/src/test/kotlin/ListComboBoxUiTest.kt
@@ -18,6 +18,7 @@ import androidx.compose.ui.test.assertHasClickAction
 import androidx.compose.ui.test.assertHasNoClickAction
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsFocused
+import androidx.compose.ui.test.assertIsNotEnabled
 import androidx.compose.ui.test.assertIsSelected
 import androidx.compose.ui.test.hasContentDescription
 import androidx.compose.ui.test.hasTestTag
@@ -242,10 +243,11 @@ class ListComboBoxUiTest {
     fun `when disabled, ComboBox cannot be interacted with`() {
         val comboBox = disabledEditableComboBox()
         comboBox.assertIsDisplayed().assertHasNoClickAction().performClick()
+        popupMenu.assertDoesNotExist()
 
-        // Ivan: It would be nice to check the absence of the OnClick,
-        // but I believe the BasicTextField adds it even on a disable state 🤷
-        textField.assertIsDisplayed().performClick()
+        // BasicTextField clickable adds an onClick action even when BTF is disabled
+        // textField.assertIsDisplayed().assertIsNotEnabled().assertHasNoClickAction().performClick() ❌
+        textField.assertIsDisplayed().assertIsNotEnabled().performClick()
         popupMenu.assertDoesNotExist()
     }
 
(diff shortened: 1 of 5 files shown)


hamen:
I had to run this #46a9b67 on Mac to satisfy the linter. This is so annoying.

ui-tests/src/test/kotlin/ListComboBoxUiTest.kt
```diff
@@ -239,14 +240,12 @@ class ListComboBoxUiTest {
...
+        val comboBox = disabledEditableComboBox()
         comboBox.assertIsDisplayed().assertHasNoClickAction().performClick()
-        composeRule
-            .onNodeWithTag("Jewel.ComboBox.NonEditableText")
-            .assertIsDisplayed()
-            .assertHasNoClickAction()
-            .performClick()
+
+        // Ivan: It would be nice to check the absence of the OnClick,
+        // but I believe the BasicTextField adds it even on a disable state 🤷
```
rock3r, line 247: >I believe the BasicTextField adds it even on a disable state

If this is true and the clickable is not enabled = false, that is a problem and it needs to be reported

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -166,30 +166,30 @@ public fun EditableComboBox(
...
 
                 Chevron(
                     isEnabled = isEnabled,
                     style = style,
                     interactionSource = interactionSource,
-                    onHoveredChanged = { chevronHovered = it },
+                    onHoveredChange = { chevronHovered = it },
                     setPopupExpanded = { popupExpanded = it },
-                    onPressWhenEnabled = onPressWhenEnabled,
+                    onPressWhenEnable = onPressWhenEnabled,
```
rock3r, line 192: I think this change was... a bit over-eager :D `onPressWhenEnabled` is correct. I can also take `onPress` (the _when enabled_ bit can be omitted)

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -217,58 +217,59 @@ public fun EditableComboBox(
...
     popupExpanded: Boolean,
     textStyle: TextStyle,
     textFieldInteractionSource: MutableInteractionSource,
     onArrowDownPress: () -> Unit,
     onArrowUpPress: () -> Unit,
     onEnterPress: () -> Unit,
-    onSetPopupExpanded: (Boolean) -> Unit,
-    onFocusedChanged: (Boolean) -> Unit,
-    onHoveredChanged: (Boolean) -> Unit,
+    onSetPopupExpandedEvent: (Boolean) -> Unit,
```
rock3r, line 231: Why `event`? It's not a common naming pattern (e.g., you don't call other lambdas `onFocusedChangeEvent` or `onClickEvent`)

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -217,58 +217,59 @@ public fun EditableComboBox(
 
 @Composable
 private fun TextInput(
-    modifier: Modifier,
```
rock3r, line 220: This was up there and without a default value on purpose...

rock3r (commented):
Apart from a bunch of unrelated/unneeded renames that should be undone, I don't understand where the "fix" actually is — it feels like a lot of non-functional changes

ui-tests/src/test/kotlin/ListComboBoxUiTest.kt
hamen: I want to be sure we are holding it correctly, and if we need, I'll open a ticket to the interested team ⌛

PR #642, opened 2024-10-17
Remove deprecated Kotlin build flag

rock3r:
`kotlin.incremental.useClasspathSnapshot` was deprecated in Kotlin 1.9, and we should not be using it anymore. This removes it.


hamen (approved):
I saw the warning yesterday while building. Perfect timing 💪

PR #644, opened 2024-10-18
Add enabled parameter to Tooltip

rock3r:
This allows the user to suppress tooltips in some circumstances. It is implemented with the if in the TooltipArea's tooltip composable since it preserves focus on the content() when switching it on/off.

Screen Recording 2024-10-18 at 19 25 45


hamen:
Tested on Windows as well. Nothing to report.

20241021-0921-20 4475223-ezgif com-video-to-gif-converter

hamen (approved):

PR #648, opened 2024-10-19
Fix tab icons not existing anymore in 243

rock3r:
Also adds tabs to the ComponentShowcaseTab IDE sample.

Not sure how this even built so far on 243...


rock3r:
This was an artifact of having switched branches to 241. The icons do exist in 242-3 :) Closing as unnecesary.

PR #652, opened 2024-10-21
Update IJP to 243 EAP 6 (243.20847.40)

rock3r:
First EAP that includes the LaFFlowService we need. Once this is merged we can build 0.26.0


hamen (approved):
Godspeed

Issue #222, opened 2023-10-31
Create ComboBox component


rock3r:
A combobox is a dropdown control that also allows you to type in a value. An example is the font size picker in the IDE Appearance settings:

<img width="168" alt="image" src="https://github.com/JetBrains/jewel/assets/153802/92efbe72-2986-420c-b187-f8165e388ce0">

As far as I can tell, Swing doesn't pre-select matching elements in the dropdown and doesn't offer auto-complete — except sometimes it does pre-select matching elements:

<img width="184" alt="image" src="https://github.com/JetBrains/jewel/assets/153802/ad91542e-1af0-447b-a38c-ca1babdad333">

(on the same page)

We should offer the ability to pre-select one element. Only single-item selections are allowed. We should also allow the user to set a validation outline.

PR #659, opened 2024-10-23
Fix ListComboBox hover

### LINKED ISSUE #631: Add hover management to ListComboBox
- [x] On hover, list item change background, but the selected item stays unchanged
- [x] On hover, clicking an item selects the item and closes the popup
- [x] On hover, ⬆️⬇️ keys change the selected item (and the background as consequence), starting from the hovered item

20241015-1602-38 3789015-ezgif com-video-to-gif-converter

Relates to #627

hamen:
Closes #631 
Relates to #222

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
@@ -48,8 +48,8 @@ public fun ListComboBox(
     val scrollState = rememberSelectableLazyListState()
     var selectedItem by remember { mutableIntStateOf(0) }
     var isListHovered by remember { mutableStateOf(false) }
-    var hoverItemIndex: Int? by remember { mutableStateOf(null) }
-    var lastHoveredIndex by remember { mutableIntStateOf(-1) }
+    var hoverItemIndex by remember { mutableStateOf(-1) }
+    var lastHoveredIndex by remember { mutableStateOf(-1) }
     var previewSelection by remember { mutableStateOf(false) }
     val scope = rememberCoroutineScope()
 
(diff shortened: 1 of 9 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -124,6 +129,8 @@ public fun ListComboBox(
                                             isItemHovered = it
                                             if (isItemHovered) {
                                                 hoverItemIndex = items.indexOf(item)
+                                                lastHoveredIndex = items.indexOf(item)
```
rock3r, line 132: Maybe I am missing something but it looks like `hoverItemIndex` and `lastHoveredIndex` always have the same value...

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -45,6 +47,8 @@ public fun ListComboBox(
     var selectedItem by remember { mutableIntStateOf(0) }
     var isListHovered by remember { mutableStateOf(false) }
     var hoverItemIndex by remember { mutableStateOf(-1) }
+    var lastHoveredIndex by remember { mutableStateOf(-1) }
+    var previewSelection by remember { mutableStateOf(false) }
```
rock3r, line 51: You don't need this if you make the hovered index nullable and just set it to null when hover is not happening

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -45,6 +47,8 @@ public fun ListComboBox(
     var selectedItem by remember { mutableIntStateOf(0) }
     var isListHovered by remember { mutableStateOf(false) }
     var hoverItemIndex by remember { mutableStateOf(-1) }
+    var lastHoveredIndex by remember { mutableStateOf(-1) }
```
rock3r, line 50:
```suggestion
    var hoverItemIndex by remember { mutableIntStateOf(-1) }
    var lastHoveredIndex by remember { mutableIntStateOf(-1) }
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
```diff
@@ -55,6 +60,6 @@ public fun SimpleListItem(
 @GenerateDataFunctions
 public class ListItemState(
     public val isSelected: Boolean,
-    public val isSoftSelected: Boolean,
     public val isHovered: Boolean,
+    public val previewSelection: Boolean,
```
rock3r, line 64: Does this need to exist? It will be true if and only if `isHovered` is true

rock3r (approved):

rock3r:
Good to ship behaviour-wise, waiting for the comments and conflicts to be addressed

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
hamen: `hoverItemIndex` is now nullable

PR #660, opened 2024-10-24
Use transparent background for disabled TextField/TextArea

### LINKED ISSUE #629: TextField should not draw its background when disabled/not editable in 242+
It looks like the platform changed how the background and borders are drawn in 242, and probably the value of `TextField.disabledBackground` too. Now, if the field is not enabled or editable, the background is simply not drawn.

We need to update our code accordingly, because we're drawing it and it looks wrong.

This seems to only be a problem in the bridge, because we use values in standalone that mask the issue, but a better solution is needed.

rosejr:
Fixes #629


rock3r (approved):
I suspect this may require an equivalent fix in standalone, but I'll look into it so it can make today's release.

PR #667, opened 2024-10-25
Fix SplitLayout divider positioning error

### LINKED ISSUE #589: SplitLayout jumps weirdly in the IDE
As we were upgrading Jewel in Studio to 0.24.2 with @jrlogsdon, we noticed that the split layout has a bad bug when dragging on the initial state. We can repro this in both Linux and macOS — have not tried Windows, but would assume it's the same. We have not tried repro'ing in standalone.

This code:

```kotlin
Column(
  modifier = Modifier.fillMaxWidth().fillMaxHeight(),
  verticalArrangement = Arrangement.Center,
  horizontalAlignment = Alignment.CenterHorizontally
) {
  HorizontalSplitLayout(
    firstPaneMinWidth = 200.dp,
    secondPaneMinWidth = 250.dp,
    first = { },
    second = { },
    state = rememberSplitLayoutState(.3f)
  )
}
```

(simplified version of [this Studio code](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:profilers-ui/src/com/android/tools/profilers/taskbased/tabs/home/TaskHomeTabComponent.kt?q=TaskHomeTab&ss=android-studio))

Has this behaviour when set as content of a toolwindow:

https://github.com/user-attachments/assets/27521866-dfa4-481f-b591-7f274890e45e

hamen:
Closes #589


samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
```diff
@@ -27,6 +27,8 @@ internal class JewelDemoToolWindowFactory : ToolWindowFactory, DumbAware {
         toolWindow.addSwingTab(SwingComparisonTabPanel(), "Swing Comparison")
 
         toolWindow.addComposeTab("Scrollbars Sample") { ScrollbarsShowcaseTab() }
+
+        toolWindow.addComposeTab("SplitLayout Sandbox") { SplitLayoutSandbox() }
```
rock3r, line 31: You don't really need it. The Releases Demo tab already has one.

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -346,29 +343,49 @@ private fun MeasureScope.doLayout(
             (constraints.maxHeight - dividerWidth).coerceAtLeast(1)
         }
 
-    val (adjustedFirstSize, adjustedSecondSize) =
-        calculateAdjustedSizes(availableSpace, minFirstPaneSizePx, minSecondPaneSizePx)
+    val minFirstPaneSizePx = with(density) { firstPaneMinWidth.toPx() }.roundToInt()
+    val minSecondPaneSizePx = with(density) { secondPaneMinWidth.toPx() }.roundToInt()
```
rock3r, line 347:
```suggestion
    val minFirstPaneSizePx = with(density) { firstPaneMinWidth.roundToPx() }
    val minSecondPaneSizePx = with(density) { secondPaneMinWidth.roundToPx() }
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
```diff
@@ -346,29 +343,49 @@ private fun MeasureScope.doLayout(
...
 
-    val firstSize: Int = firstGap.roundToInt().coerceIn(adjustedFirstSize, availableSpace - adjustedSecondSize)
-
-    val secondSize = availableSpace - firstSize
+    // Ensure sizes respect minimum widths
+    val (adjustedFirstSize, adjustedSecondSize) =
+        calculateAdjustedSizes(
+            availableSpace,
+            initialFirstSize.coerceAtLeast(minFirstPaneSizePx),
+            initialSecondSize.coerceAtLeast(minSecondPaneSizePx),
```
rock3r, line 358: I'd expect the coerce to happen inside the `calculateAdjustedSizes()`, not here... and looking at the code of that function, it kind of does, just not very well

samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
hamen: I meant to remove it already and I did in a4778299

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
hamen: Done in https://github.com/JetBrains/jewel/pull/667/commits/422b9729673139f453f65720fc2c5247fc419fec

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SplitLayout.kt
hamen: It was indeed redundant. Thank you.

Done in https://github.com/JetBrains/jewel/pull/667/commits/a10ed6dc841518e5ea1ab85493ea29b76d6108f9

rock3r (approved):

PR #669, opened 2024-10-30
Reduce duplication in ListComboBox

### LINKED ISSUE #649: Reduce duplication in ListComboBox

hamen:
Closes #649

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
@@ -98,33 +98,31 @@ public fun ListComboBox(
         items: List<String>,
         onHoverItemChange: (String) -> Unit,
         listItemContent: @Composable (String, Boolean, Boolean, Boolean, Boolean) -> Unit,
-    ): SelectableLazyListScope.() -> Unit {
-        return {
-            items(
-                items = items,
-                itemContent = { item ->
-                    var isItemHovered by remember { mutableStateOf(false) }
-                    Box(
-                        modifier =
-                            Modifier.onHover {
-                                isItemHovered = it
-                                if (isItemHovered) {
-                                    hoverItemIndex = items.indexOf(item)
-                                    onHoverItemChange(item)
-                                }
+    ): SelectableLazyListScope.() -> Unit = {
+        items(
+            items = items,
+            itemContent = { item ->
+                var isItemHovered by remember { mutableStateOf(false) }
+                Box(
+                    modifier =
+                        Modifier.onHover {
+                            isItemHovered = it
+                            if (isItemHovered) {
+                                hoverItemIndex = items.indexOf(item)
+                                onHoverItemChange(item)
                             }
-                    ) {
-                        listItemContent(
-                            item,
-                            isSelected,
-                            isActive,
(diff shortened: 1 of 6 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -76,6 +78,82 @@ public fun ListComboBox(
...
+            onSelectedItemChange(items[selectedItemIndex])
+        }
+    }
+
+    fun contentItems(
+        items: List<String>,
+        onHoverItemChange: (String) -> Unit,
+        listItemContent: @Composable (String, Boolean, Boolean, Boolean, Boolean) -> Unit,
+    ): SelectableLazyListScope.() -> Unit {
+        return {
```
rock3r, line 102: Nit: this reads weird. Why not using expression syntax?

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -76,6 +78,82 @@ public fun ListComboBox(
...
+            scrollState = scrollState.lazyListState,
+            modifier =
+                Modifier.heightIn(max = popupMaxHeight).onHover {
+                    isListHovered = it
+                    onListHoverChange(it)
+                },
+        ) {
+            SelectableLazyColumn(
+                modifier =
+                    Modifier.fillMaxWidth().heightIn(max = popupMaxHeight).padding(horizontal = 6.dp, vertical = 6.dp),
```
rock3r, line 149: I'm aware this was not introduced here, but are you sure this padding should not be a contentPadding instead? And, should it be in the styling?

```suggestion
                    Modifier.fillMaxWidth().heightIn(max = popupMaxHeight).padding(6.dp),
```

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
hamen: Done in https://github.com/JetBrains/jewel/pull/669/commits/babd76c1b046aac530cf819fde56c9eb7be389f6

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
hamen: I'm in favor of extracting it into the style: https://github.com/JetBrains/jewel/pull/669/commits/edf0314170214d6416e1dcbe891d355354743f17

The problem that I have now comes from the ListComboBox being a bit “arbitrary” when it comes to metrics because all the comboboxes in UI Sandbox, Preferences, Figma are different: I can't find that `6.dp` we are using in the IntelliJ source code 😬
Any suggestion is welcome 🙏🏻

rock3r: As usual, my approach is at a high level to try and get to this 6.dp from values we can obtain from the LaF; however, if they don't exist in the LaF, and these are hardcoded in Swing, even as sum of other hardcoded values, it's ok to hardcode them on our side too.

If they come in any way shape or form from the LaF in Swing, then they need to be constructed from the same values (e.g., a LaF value plus a fixed value)

Issue #691, opened 2024-11-18
ListComboBox arrow keys behaviour is inconsistent with Swing


hamen:
One small nit on behavior is with arrow keys. When using the Up Arrow to navigate through the list to the top, once you get to the top, the subsequent key press moves the insertion cursor in the `TextField` to the beginning.
![arrowing-up](https://github.com/user-attachments/assets/e008a8a9-6dfa-4608-980c-ad388bc9690a)

It's likely behavior inherited from using `TextField`, but looking at the Swing implementation, it doesn't do this. So curious if we should be consistent here.

_Originally posted by @c5inco in https://github.com/JetBrains/jewel/issues/627#issuecomment-2480110952_

hamen:
Closed with #692

PR #692, opened 2024-11-18
In ListComboBox prevents caret from moving on key up/down pressed

hamen:
Relates to #691 

Recording2024-11-18114501-ezgif com-video-to-gif-converter

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
@@ -247,11 +247,9 @@ private fun TextInput(
                 .onFocusChanged { onFocusedChange(it.isFocused) }
                 .focusRequester(textFieldFocusRequester)
                 .onPreviewKeyEvent {
-                    // We are consuming the event to prevent the caret
-                    // from moving to the start and to the end of the text field
-                    // when the user presses the arrow up/down keys
+                    if (it.type != KeyEventType.KeyDown) return@onPreviewKeyEvent false
                     when {
-                        it.type == KeyEventType.KeyDown && it.key == Key.DirectionDown -> {
+                        it.key == Key.DirectionDown -> {
                             if (popupExpanded) {
                                 onArrowDownPress()
                             } else {
@@ -260,7 +258,7 @@ private fun TextInput(
                             }
                             true
                         }
-                        it.type == KeyEventType.KeyDown && it.key == Key.DirectionUp -> {
+                        it.key == Key.DirectionUp -> {
                             if (popupExpanded) {
                                 onArrowUpPress()
                             } else {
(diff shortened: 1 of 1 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -247,32 +247,40 @@ private fun TextInput(
...
-                        if (popupExpanded) {
-                            onArrowDownPress()
-                        } else {
-                            onSetPopupExpanded(true)
-                            textFieldFocusRequester.requestFocus()
+                    // We are consuming the event to prevent the caret
+                    // from moving to the start and to the end of the text field
+                    // when the user presses the arrow up/down keys
+                    when {
+                        it.type == KeyEventType.KeyDown && it.key == Key.DirectionDown -> {
```
rock3r, line 254: Nit: you can extract the type check to an if that quick-returns false if it's not KeyDown, before the when, simplifying the when conditions (which all check that)

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -247,32 +247,40 @@ private fun TextInput(
...
-                    }
-
-                    if (it.type == KeyEventType.KeyDown && it.key == Key.Enter) {
-                        if (popupExpanded) {
+                        it.type == KeyEventType.KeyDown && it.key == Key.Enter -> {
+                            if (popupExpanded) {
+                                onSetPopupExpanded(false)
+                            }
+                            onEnterPress()
+                            false
```
rock3r, line 276: I know it was already essentially doing this, but shouldn't this be `true` given we're consuming the enter press?

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -247,32 +247,40 @@ private fun TextInput(
...
+                        }
+                        it.type == KeyEventType.KeyDown && it.key == Key.Escape && popupExpanded -> {
                             onSetPopupExpanded(false)
+                            false
```
rock3r, line 280: Ditto

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -247,32 +247,40 @@ private fun TextInput(
...
                 .onPreviewKeyEvent {
-                    if (it.type == KeyEventType.KeyDown && it.key == Key.DirectionDown) {
-                        if (popupExpanded) {
-                            onArrowDownPress()
-                        } else {
-                            onSetPopupExpanded(true)
-                            textFieldFocusRequester.requestFocus()
+                    // We are consuming the event to prevent the caret
+                    // from moving to the start and to the end of the text field
+                    // when the user presses the arrow up/down keys
```
rock3r, line 252: This comment here was misleading for me — it's related to L261 and L269 only, but placed far from either; also, other branches in the when should return `true` as well...

hamen: 💯 Removed in https://github.com/JetBrains/jewel/pull/692/commits/05adbe92c586e327936c5d86a8021d3747471685

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
hamen: Done in https://github.com/JetBrains/jewel/pull/692/commits/05adbe92c586e327936c5d86a8021d3747471685

PR #694, opened 2024-11-19
Markdown: synchronize scrolling between editor and preview

AlexVanGogen:
Small demo of how it works in IntelliJ (source scrolling events are passed to the syncronizer with a debounce timeout of 16ms):

https://github.com/user-attachments/assets/853ecbef-de9c-4c6f-a8b1-0b4ffedec039

I'm also contemplating writing some sanity tests that validate the coherency of internal structures (which means internals of `ScrollingSynchronizer` will likely be exposed for testing).


markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -252,16 +260,19 @@ public open class DefaultMarkdownBlockRenderer(
...
-                        text = "$number${block.delimiter}",
-                        style = styling.numberStyle,
-                        color = styling.numberStyle.color.takeOrElse { LocalContentColor.current },
-                        modifier =
-                            Modifier.focusProperties { canFocus = false }
-                                .widthIn(min = styling.numberMinWidth)
-                                .pointerHoverIcon(PointerIcon.Default, overrideDescendants = true),
-                        textAlign = styling.numberTextAlign,
-                    )
+                    AutoScrollableBlock(block) { modifier ->
```
rock3r, line 263: I'd prefer if we had a separate renderer for this, `ScrollSyncMarkdownBlockRenderer: DefaultMarkdownBlockRenderer` that overrides these functions as needed.

This way we don't have a perf/memory penalty on all the non-scroll-synced use cases, which are way more common

markdown/extension/gfm-alerts/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/github/alerts/GitHubAlertBlockRenderer.kt
```diff
@@ -66,50 +66,56 @@ public class GitHubAlertBlockRenderer(private val styling: AlertStyling, private
         onUrlClick: (String) -> Unit,
         onTextClick: () -> Unit,
     ) {
-        Column(
-            Modifier.drawBehind {
-                    val isLtr = layoutDirection == Ltr
-                    val lineWidthPx = styling.lineWidth.toPx()
-                    val x = if (isLtr) lineWidthPx / 2 else size.width - lineWidthPx / 2
+        AutoScrollableBlock(block) { modifier ->
```
rock3r, line 69: Ditto here, I'd prefer a separate renderer for this use case

rock3r: I'm ok with this. Also alerts tend to be small so just being able to reach them at the block level is probably absolutely fine

rock3r:
Superseeded by https://github.com/JetBrains/intellij-community/pull/2908

PR #698, opened 2024-11-25
Finalize text message and Actions slot API

hamen:
image

### DIFF
--- a/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Banner.kt
+++ b/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Banner.kt
@@ -80,7 +80,7 @@ private fun BannerImpl(
     actions: (@Composable RowScope.() -> Unit)?,
     modifier: Modifier = Modifier,
 ) {
-    Column(modifier = modifier.height(40.dp).background(style.colors.background),) {
+    Column(modifier = modifier.height(40.dp).background(style.colors.background)) {
         Divider(orientation = Orientation.Horizontal, color = style.colors.border)
         Row(
             verticalAlignment = Alignment.CenterVertically,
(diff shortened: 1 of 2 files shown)


ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Banner.kt
```diff
@@ -78,18 +80,26 @@ private fun BannerImpl(
     actions: (@Composable RowScope.() -> Unit)?,
     modifier: Modifier = Modifier,
 ) {
-    Column(modifier = modifier) {
+    Column(
+        modifier = modifier.height(40.dp).background(style.colors.background),
+        verticalArrangement = Arrangement.Center,
```
rock3r, line 85: I don't think this does what you think it does :D And I'd argue this is wrong to have

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Banner.kt
```diff
@@ -78,18 +80,26 @@ private fun BannerImpl(
...
-            Spacer(modifier = Modifier.weight(1f))
+            Text(
+                text = text,
+                style = textStyle,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+                modifier = Modifier.weight(1f),
+            )
             if (actions != null) {
                 Row(
```
rock3r, line 104: I think we need to add a Spacer before this Row, so there is room between the text and the actions

rock3r: Inside the if, of course :)

hamen: No spacer
image

Spacer
image

I'm adding the Spacer

hamen: Done in https://github.com/JetBrains/jewel/pull/698/commits/0e16598c6b0377735fd4b174998edbf623e1f112

ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Banner.kt
hamen: Done in https://github.com/JetBrains/jewel/pull/698/commits/0ae8318434bbed30eb918f713ba772d2c2c80d9d

rock3r (approved):

PR #704, opened 2024-11-27
Tune logging

rock3r:
This PR uncomments logging in `ResourcePainterProvider` which should never have been committed to begin with, and adjusts its level to trace (FINER). It also sets the `JavaFactory` level to `FINE` (debug), so that `ResourcePainterProvider`'s logs don't spam the console. Lastly, it cleans up `JewelLogger`'s code with the IDE's _Clean up_ action


hamen (approved):
Sorry for having committed this. It was a mistake: usually I mark the file as “assume unchanged”, but it slipped through.

PR #705, opened 2024-11-27
Fix indeterminate state not working in Checkboxes

rock3r:
The issue was that we weren't considering indeterminate as selected, icon-wise, but the IDE resources do consider it as such.


hamen (approved):
Nice catch 👍

PR #717, opened 2024-12-06
Fix disabled state dropdown menu item color in IDE

### LINKED ISSUE #668: Disabled options text color is incorrect in `Dropdown` component
I am an engineer for Android Studio Profiler which, as of Koala Feature Drop, has been using Jewel for the home tab implementation. Recently we had the following bug reported: https://b.corp.google.com/issues/376056212. This was also reproduced in another area of Studio that employs Jewel (`Device Manager` > `Add a new device`). This issue first appeared in Ladybug Feature Drop, and was not present in Koala Feature Drop nor Ladybug 1.

Essentially, the issue can be summarized as, when a single disabled dropdown option, or even if the dropdown itself is disabled, the text color shown makes it look like its enabled.

To reproduce, I simply cloned the jewel project, set `enabled` equal to `false` for the only `Dropdown` in the `ComposeShowcaseTab` and ran the IDE Sample. Here is the result:
image

hamen:
Fix #668 
a bit more 😀🤞


rock3r (approved):

hamen:
Migrated to IJP as https://github.com/JetBrains/intellij-community/pull/2904

Issue #491, opened 2024-07-23
Re-evaluate whether we need standalone styling default factories to be composable


rock3r:
Without text styles, after #490, it should not be needed anymore

PR #721, opened 2024-12-10
Add RTL Support for Chevron Icons in LazyTreeIcons Defaults

kdroidFilter:
This pull request adds support for right-to-left (RTL) layout direction in the LazyTreeIcons.Companion.defaults function of the Jewel theme for JetBrains. Previously, the chevron icons used in the tree structure did not account for layout direction, leading to incorrect icon orientation in RTL contexts. This update ensures that the chevron icons adjust dynamically based on the current layout direction.


kdroidFilter:
@rock3r Ok, I'm trying to do this tonight

rock3r:
Much appreciated, thanks!

rock3r:
Thanks for updating the PR.

I've looked at the change, and I am torn on it — for these reasons:
 * This makes a style creation composable when we're actively trying to make them not (see #491)
 * We don't really do anything explicit to support RTL layouts — not because we don't want to, but because the Int UI design system does not, neither in the specs nor in the IDEs. This means that what the desired behaviour actually is remains largely undefined, and this would be the only component with any such logic built, which is inconsistent
 * You do not need to change Jewel to do this; users can provide their own tree styling with a `CompositionLocalProvider` that has the icon they prefer/need, using the existing APIs Jewel offers

As a result, I'm leaning towards not taking this in. @hamen what do you think?

kdroidFilter:
@rock3r I understand that the goal is to avoid making styles composable. My intention was simply to ensure that the icons adapt dynamically to the RTL context.

I must admit that I didn’t realize the theme was exclusively intended for use in IDEs. I thought it was also designed for building desktop applications more generally. Apologies for the inconvenience caused by this misunderstanding.

hamen:
As much I appreciate external contributions, right now, I would focus on providing the core tools that mirror what the IJP and Swing are doing.
In the future, once we reach “feature parity”, we can think of iterating on this more, and expand the platform and library possibilities.
Considering your scenario, I'm glad that you can achieve your goal working “around” Jewel. This proves that the APIs are flexible enough to be customized, still providing a set of convenient defaults.
I would keep this PR closed as a reminder of future possible expansions.

rock3r:
Makes sense to me. I'll close this PR then, but still want to thank you for the contribution even if it wasn't merged @kdroidFilter 🙏

PR #723, opened 2024-12-10
Fix excessive recompositions in SelectableLazyColumn

rock3r:
The code that reacts to selected keys change was triggering too often and causing issues with excessive recompositions, which could cascade into weird scrolling/state management issues.


hamen (approved):

hamen:
Migrate to IJP as https://github.com/JetBrains/intellij-community/pull/2905

PR #727, opened 2024-12-11
Add slot-based SimpleListItem overloads

rock3r:
This lets folks define more complex list item content layouts (e.g., multiple text runs, additional icons, etc). For example, one may use it to build items like these:

Completion menu screenshot


hamen (approved):
It makes super sense. Nice iteration.

hamen:
Migrated to IJP as https://github.com/JetBrains/intellij-community/pull/2912

PR #728, opened 2024-12-12
Standalone hot reload

rock3r:
Adds Hot Reload to the standalone project.

This requires bumping to a forked Kotlin 2.1.0, and Kotlinter 5.0.0 (with ktlint 1.5.0)


rock3r:
Closing as this would need to be ported to the `intellij-community` repo, but it'll be a whole process :)

Issue #729, opened 2024-12-13
Menu for DecoratedWindow


MFlisar:
The decorated window does not support the default menu - IntelliJ does have something as replacement as you can see in the following screenshots.

Is something like this also possible with this library? I think not... Please let me know if it is possible or consider this as a feature request.

**Menu**

Clicking the menu icon with the 3 lines opens the menu inside the decorated title bar.

![grafik](https://github.com/user-attachments/assets/c61cf5ff-6d94-45bd-822b-77bbc699d64c)

![grafik](https://github.com/user-attachments/assets/23446088-f073-49ff-a139-e00a2f70449c)

rock3r:
You can put whatever you want in the window decoration, including buttons that have a text label and when clicked open a PopupMenu. We're not planning to implement this further for now, but contributions are welcome as usual

MFlisar:
That's what I'm doing right now, but it behaves a little buggy... `PopupMenu` does not open immediately if it is placed inside the `TitleBar` - I always must move the mouse after clicking and only afterwards the popup opens.

Code looks like following:

```kotlin
TitleBar(Modifier.newFullscreenControls()) {
    Row(
        modifier = Modifier.align(Alignment.Start),
        verticalAlignment = Alignment.CenterVertically
    ) {
        Dropdown(
            modifier = Modifier.height(30.dp),
            menuContent = {
                selectableItem(
                    selected = false,
                    onClick = { }
                ) {
                    TitleMenuItem("Sub Menu", null)
                }
            }
        ) {
            TitleMenuItem("Test Menu", null, false)
        }
    }
}

@Composable
private fun TitleMenuItem(title: String, imageVector: ImageVector?, popup: Boolean = true) {

    val jewelTitleTheme = org.jetbrains.jewel.foundation.theme.JewelTheme.defaultTitleBarStyle
    val contentColor = org.jetbrains.jewel.foundation.theme.JewelTheme.contentColor
    val foreground = if (popup) contentColor else jewelTitleTheme.colors.content

    Row(
        horizontalArrangement = Arrangement.spacedBy(4.dp),
        verticalAlignment = Alignment.CenterVertically,
    ) {
        if (imageVector != null) {
            Icon(
                modifier = Modifier.size(30.dp).padding(4.dp),
                imageVector = imageVector,
                contentDescription = null,
                tint = foreground
            )
        } else {
            if (popup) {
                Spacer(modifier = Modifier.size(30.dp).padding(4.dp))
            }
        }
[...]

Issue #424, opened 2024-07-02
Welcome menu glitch on Windows 11


hamen:
On Windows 11 the Welcome menu stays closed even if you click on it. It opens as soon as your mouse moves by 1 pixel in any direction.

https://github.com/JetBrains/jewel/assets/19003/1f51898d-1b9e-4133-97a9-22c26b8ae677

Issue #368, opened 2024-04-26
Dragging on the TitleBar to move the window does not always work 


hvisser:
Dragging the TitleBar doesn't always work, at least on macOS with jbr-17 and jbr-21. When I use a normal `Window` the window can be dragged normally.  Since TitleBar is delegating the hit test to the JetBrains runtime, this might also be a runtime bug.

https://github.com/JetBrains/jewel/assets/866834/e504719b-d70e-46ba-bb4e-3d2522920b78

hvisser:
I've also tested this with an completely empty window, so it doesn't seem to be related to focus I think.

rock3r:
Thanks :) @devkanro have you ever seen this?

devkanro:
This is a known issue, and I have no idea currently, will look into this, but I think it is a JBR issue.

hvisser:
The pointer is handled here https://github.com/JetBrains/jewel/blob/main/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.Windows.kt#L43 and it looks OK, but maybe the actual pointer event does not always get delivered to the window decoration correctly?

FWIW I assume that Intellij and other products do something similar but using Swing API's and I'm not observing that issue with those products.

hmy65:
pretty sure this issue existed on windows as well, using latest 243 with jbr21, still sadly existed :（

rock3r:
@devkanro any chance you could look into this?

yaroslavkulinich:
This issue significantly impacts usability, and I hope this gets prioritized for a fix soon.

rock3r:
I hope @devkanro will come back to help with this; they're the only person who's got knowledge on the topic

hmy65:
@rock3r @devkanro 
It's weird Intellij does not have this issue at all. So part of the implementation must be close source.

MFlisar:
Here's my input to this issue:

The `TitleBar` is on top of it's content inside the `composable` tree - events are propagated from top to bottom. `titleBar.forceHitTest` does only have an effect on the **next** event as far as I can see in my tests.... I think, that you can't intercept the events of child from somewhere above in the hierarchy without a fully custom event handling if the implementation of the title bar does not work with the consumed flags but by calling a function *after* the event has reached the `TitleBar` already...

**Suggestion**

[...]

PR #731, opened 2024-12-15
Bugfix decorated window on windows - clickable components inside title area should work now

MFlisar:
**Idea**

Place a background composable behind the title content that handles all windows draggable header related stuff. This way, everything inside the title bar is placed **on top** of the composable that forwards the events to the draggable header.

**Issues this solves**

when clicking a dropdown menu in the title bar, the popup did not pop up until the next mouse move => fixed

**TODO**

* the macos implementation does use `customTitleBarMouseEventHandler` as well - it may be necessary to use the same fix there? I don't know, I don't have a mac and I don't know if the mac has the same issues
* for the sake of minimal adjustments I kept the `customTitleBarMouseEventHandler` as is (also because I don't know how it is used on mac) - it may be possible to simplifiy it like following as there is no need to try to detect if we are over a user control or not anymore because of the layerig of the background composable and the content.

```
internal fun Modifier.customTitleBarMouseEventHandler(titleBar: CustomTitleBar): Modifier =
    pointerInput(Unit) {
        val currentContext = currentCoroutineContext()
        awaitPointerEventScope {
            while (currentContext.isActive) {
                awaitPointerEvent(PointerEventPass.Main)
                titleBar.forceHitTest(false)
            }
        }
    }
```

### DIFF
--- a/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.Windows.kt
+++ b/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.Windows.kt
@@ -1,6 +1,8 @@
 package org.jetbrains.jewel.window
 
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
@@ -27,7 +29,7 @@ internal fun DecoratedWindowScope.TitleBarOnWindows(
     val titleBar = remember { JBR.getWindowDecorations().createCustomTitleBar() }
 
     TitleBarImpl(
-        modifier = modifier.customTitleBarMouseEventHandler(titleBar),
+        modifier = modifier,
         gradientStartColor = gradientStartColor,
         style = style,
         applyTitleBar = { height, _ ->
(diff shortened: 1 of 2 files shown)


rock3r:
Hi, I ported & tested the change on macOS and it doesn't seem to make any difference with the previous behaviour. Dragging the title bar sometimes works, and sometimes it doesn't. I can't figure out what the reason is... there doesn't seem to be much of a pattern. I also tried mucking around a bit with the code to see if I could get it to work, but it still has the same issue. This needs some more investigation...

rock3r:
@MFlisar I pushed the tweaks I did locally to a PR on your fork, if you want to incorporate them: https://github.com/MFlisar/jewel/pull/1

MFlisar:
Can you tell me if the click behavior on the dropdown menu is the same as for me (before and after)? If so, can you reproduce the improvement in this case?

Actually I can't see any issues with drag&drop, it behaves as it should for me when I play around for a few minutes.

rock3r:
@MFlisar the dropdown in the standalone sample works fine for me on the main branch if that's what you're asking

MFlisar:
Sorry for being so detailed, but does that mean that the popup always opens immediately without the need to move the mouse after the click for one pixel?

Because that is the issue this pull requests solves for me on windows 11, latest release version...

rock3r:
@MFlisar ah, no — I thought this was a fix for #368, but you're saying it's for #424 instead? I have been trying to repro and assess it as a fix for the former, as you started the discussion on #368

rock3r:
To be clear, #424 is only happening on Windows, not on Mac nor Linux

MFlisar:
> To be clear, #424 is only happening on Windows, not on Mac nor Linux

Yes, that's the issue... I'm sorry, I commented in the wring issue because you linked to it from my issue (https://github.com/JetBrains/jewel/issues/729) - with this fix I can use a menu that works on windows.

rock3r:
Ah, sorry for the confusion! @hamen since you reported #424, can you check if this PR solves the issue for you?

hamen:
It seems to be working on Windows 👍🏻 

Recording2024-12-17160316-ezgif com-video-to-gif-converter

PR #754, opened 2025-01-20
Clean up Jewel dialogs in IDE

rock3r:
* The components showcase dialog setup makes more sense now (started from an action instead of a random button in the toolwindow)
* Dialogs don't deadlock the IDE process hard anymore if opened before any other Compose UI is shown (workaround for IJPL-166436)
* Jewel actions have been renamed and cleaned up
* The whole components showcase modifier setup now makes sense to hoist (and uses the right padding default in standalone)
* Icon buttons now allow passing a modifier to the inner icon
* Toolbar icons in the showcase look (more) like the IDE ones now

### Standalone

image

### In-IDE

image

image


hamen:
Migrated to IJC as https://github.com/JetBrains/intellij-community/pull/2921

